PR #6, opened 2013-11-11
Adds a couple of dependencies for initial migration

andrewjesaitis:
While setting up Varify, I noticed that there were a couple of unmet dependencies for the initial migrations to cleanly take place. eed5bb7414cde802963df0fa5145d450b6068491 fixes the assessment migration that actually throws an exception. While, e19addc209e3112c52ae061a3afa3fb3e7745f79, allows the last fixture to be loaded. 

:thumbsup: Thanks for building this--it looks awesome!

### DIFF
--- a/varify/assessments/migrations/0001_initial.py
+++ b/varify/assessments/migrations/0001_initial.py
@@ -7,6 +7,10 @@ from django.db import models
 
 class Migration(SchemaMigration):
 
+    depends_on = (
+        ("samples", "0001_initial"),
+    )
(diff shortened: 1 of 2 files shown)


bruth:
Thanks so much for taking a look at this! If you need any other help or run into any other issues, don't hesitate to file more issues or PRs. You can also chat with us here: http://www.hipchat.com/gZcKr0p3y (noted on the README) during the day (eastern time in the US).

PR #23, opened 2014-01-09
Implement custom results workflow, remove analysis workflow

naegelyd:
The analysis workflow is unnecessary. The query workflow should be workable or it can be modified. This provides the variant formatted variant summaries. Still needs work on variant details and exporter modals but this was more to get a custom workflow implemented and the variant data retrieval sorted out.


varify/static/scripts/coffeescript/main.coffee
```diff
@@ -8,7 +8,7 @@ require
         tpl:
             variable: 'data'
 
-, ['cilantro', '../../scripts/javascript/src/ui'], (c, ui) ->
+, ['cilantro', '../../scripts/javascript/src/ui', 'project/csrf'], (c, ui, csrft) ->
```
bruth: Not that is matters, but the `csrft` variable is spelled wrong.

varify/static/scripts/coffeescript/ui/tables/row.coffee
```diff
@@ -0,0 +1,81 @@
...
+
+            genomicPosition = new cell.GenomicPosition
+                chromosome: variant.chr
+                position: variant.chr
+                assessment: assessment
+
+            condensedFlags = new cell.CondensedFlags
+                variant: variant
+
+            @$el.html ''
```
bruth: Using `@$el.empty()` is a tad nicer.

varify/static/scripts/coffeescript/ui/tables/row.coffee
```diff
@@ -0,0 +1,81 @@
...
+            @$el.append gene.render().el,
+                        hgvsP.render().el,
+                        variantEffects.render().el,
+                        hgvsC.render().el,
+                        genotype.render().el,
+                        genomicPosition.render().el,
+                        condensedFlags.render().el
+
+        onRender: =>
+            @model.fetch()
```
bruth, line 72: This may be a CSS thing, but right now the rows are not visible to end users until the data is synced.. which feels slow. Even if the row outline was visible (like a placeholder), it would feel faster.

bruth, line 72: Nevermind.. I see the `.variant-container` CSS class and it has a border.

bruth:
The templates defined for Varify could be a directory outside of Cilantro. The [module paths](https://github.com/cbmi/varify/pull/23/files#diff-a632d8b7914212c417b8bd34d5d123e2R16) just need to be updated to point to the other location. The `require` configuration has a set of paths defined (currently extended [here](https://github.com/cbmi/varify/blob/master/varify/templates/cilantro/config.js)) that act as aliases to paths relative to the `baseUrl`. Moving these outside of the Cilantro directory will make upgrades easier since we wouldn't have to worry about managing the custom files. However, this is just a suggestion, but not necessary.

naegelyd:
I agree that the templates should be moved but I will create a separate issue for that and handle it later.

PR #27, opened 2014-01-15
Fully replicate the result modal from the internal version

naegelyd:
I am not a fan of some parts of this but I kept it as close to the original structure of the Varify code for the sake of time. We can improve in the future but it all seems to be working for now.


mitalia:
While it's fresh in your mind, can you put some tickets in for "future" releases so we don't lose track of the stuff that we want to change.

varify/static/scripts/coffeescript/ui/workflows.coffee
```diff
@@ -1,5 +1,5 @@
 define [
-    'underscore'
+    'underscore',
```
bruth: Harmless, but no need for this comma.

varify/variants/resources.py
```diff
@@ -63,63 +79,119 @@ def is_not_found(self, request, response, pk):
...
+            # use this later to make sure that we don't expose the assessment
+            # details made via samples this user doesn't have rights to view.
+            user_project_ids = \
+                get_objects_for_user(request.user, 'samples.view_project') \
+                .values_list('pk', flat=True)
+
+            data['assessments'] = []
+
+            assessment_data = {}
+            for assessment in assessments:
```
bruth: This iteration confused me. The `assessment_data` variable is declared outside the loop, but is only used within the loop. You can also shorten the variable name in the loop so you don't need to escape the newline:

``` python
for assessment in assessments:
    data = {
        'id': assessment.pk,
        'pathogenicity': assessment.pathogenicity.name,
        # ...
    }
```

bruth: I guess don't call it `data` since that is in the outer scope.. but something shorter.

bruth:
I presume most of the code is reused? Nothing stood out as being incorrect.

naegelyd:
Correct. For the most part, this code is largely a copy of the internal code base with some changes to use Marionette.ItemView over Backbone.View and minor styling differences.

PR #43, opened 2014-01-23
Add column headers and enable sorting on results table

### LINKED ISSUE #30: Add column headers to variant summary table
These headers(as they are in Cilantro) should be clickable to enable sorting on that column. This might be difficult because some of the columns pull in elements from multiple concepts/fields but we can just go with the dominant(first?) element in the column to use for sorting.

naegelyd:
Fixes #30.


varify/static/scripts/coffeescript/ui/tables/header.coffee
```diff
@@ -0,0 +1,78 @@
...
+
+            model = _.find @data.view.facets.models, (f) ->
+                return f.get('concept') == concept
+
+            # If this column is not in the view already, add it in before
+            # updating the view sort properties.
+            if not model?
+                @data.view.facets.add({concept: concept})
+
+            console.log("Clicked #{ concept }")
```
bruth: Remove logging

varify/static/stylesheets/scss/style.scss
```diff
@@ -9,9 +9,18 @@ body {
     margin-top: 10px;
 }
 
+.unsortable-column {
+    cursor: auto;
+    pointer-events: none;
```
bruth, line 14: Neat.. I was not aware of this feature. Looks like it is not supported in IE <11, but I assume that should be harmless?

naegelyd, line 14: This will break the sorting for IE < 11 because the click events will be registering on the `<span>...</span>` or `<i>...</i>` in the column headers instead of the `<th>...</th>` elements. Since we can't assume everyone will be on IE11 or another supported browser version, I will do this in a different way. It doesn't look like IE 11 adaption is as widespread as I had thought...

bruth, line 14: Could you do a `event.preventDefault()` in the view? That usually feels more clean and obvious since it doesn't spread out the behavior across javascript and CSS. Or at least do `$el.css('pointer-events', 'none')` or something in addition to preventing the behavior (although that may be redundant).

PR #46, opened 2014-01-24
Add result score model and migration

naegelyd:


mitalia:
After discussion with Aaron, we will add a rank as well. Rank will be the primary sort column in the UI. We should display the the score for information purposes.

PR #47, opened 2014-01-24
Add variant score column to results table

### LINKED ISSUE #42: Add variant score column to results table
A column should be added to the condensed view to display the phenotype score from the scoring service.

naegelyd:
Fixes #42.


varify/static/scripts/coffeescript/templates.coffee
```diff
@@ -9,6 +9,15 @@ define [
         pchr = Numbers.toDelimitedNumber(pos)
         return "<td class=genomic-position><a target=_blank href=\"http://genome.ucsc.edu/cgi-bin/hgTracks?position=chr#{ chr }%3A#{ pos }\">chr#{ chr } <span class=muted>@</span> #{ pchr }</a></td>"
 
+    phenotypeScore = (resultScore) ->
+        if resultScore? and resultScore.rank? and resultScore.score?
+            rank = parseInt(resultScore.rank)
+            score = parseFloat(resultScore.score)
+
+            if not isNaN(rank) and not isNaN(score)
```
bruth, line 17: Since you are checking is this a NaN, the `resultScore.rank?` and `resultScore.score?` above are redundant since `parse*` will return a NaN for null or undefined values.

PR #63, opened 2014-02-03
Use log statements over print/sys.stdout

### LINKED ISSUE #55: Convert print statements to logger calls
Statements like [these](https://github.com/cbmi/varify/blob/master/varify/samples/management/subcommands/queue.py#L59-L64) should be refactored to use logging calls rather than print calls. This is a problem across many files(especially the commands) so the code linked above is only an example to highlight the problem this issue is meant to resolve. I think all current print statements can be converted to log statements and if not, the reason for the print statement usage should be noted as a line comment.

naegelyd:
Fixes #55.

This also contains formatting and style cleanups.

### DIFF
--- a/varify/genes/management/subcommands/load.py
+++ b/varify/genes/management/subcommands/load.py
@@ -1,15 +1,13 @@
-import re
-import sys
 import logging
-import traceback
+import re
 from optparse import make_option
 try:
     from collections import OrderedDict
 except ImportError:
     from ordereddict import OrderedDict
-from django.db import transaction, connections, DEFAULT_DB_ALIAS
+from django.db import connections, DEFAULT_DB_ALIAS
 from django.core.management.base import BaseCommand
-from varify.genes.models import GeneFamily, Gene, Synonym, GenePhenotype
+from varify.genes.models import GeneFamily, Gene, Synonym
 from varify.genome.models import Chromosome
 from varify.literature.models import PubMed
 
@@ -17,17 +15,18 @@ log = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    """Cleans and loads genes from various sources into a uniform set of
-    tables.
+    """
+    Cleans and loads genes from various sources into a uniform set of tables.
     """
     option_list = BaseCommand.option_list + (
         make_option('--database', action='store', dest='database',
-            default=DEFAULT_DB_ALIAS, help='Nominates a database to print the '
-                'SQL for.  Defaults to the "default" database.'),
+                    default=DEFAULT_DB_ALIAS,
+                    help='Nominates a database to print the SQL for. Defaults '
+                         'to the "default" database.'),
         make_option('--genes', action='store_true', default=False,
-            help='Reload genes, synonyms and PubMed IDs from HGNC'),
+                    help='Reload genes, synonyms and PubMed IDs from HGNC'),
         make_option('--families', action='store_true', default=False,
-            help='Reload gene families provided by HGNC'),
+                    help='Reload gene families provided by HGNC'),
     )
 
     def _get_or_create_gene(self, record):
@@ -44,10 +43,11 @@ class Command(BaseCommand):
         else:
             match = self.chrom_re.match(record['chromosome'])
             if not match:
-                log.warning('unable to match gene chromosome from HGNC', extra={
-                    'hgnc_id': record['hgnc_id'],
-                    'raw_chr': record['chromosome'],
-                })
+                log.warning('unable to match gene chromosome from HGNC',
+                            extra={
+                                'hgnc_id': record['hgnc_id'],
+                                'raw_chr': record['chromosome'],
+                            })
                 return None, False
             target['chr_id'] = self.chromosomes[match.groups()[0]]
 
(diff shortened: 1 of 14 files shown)


varify/genes/management/subcommands/load.py
```diff
@@ -154,8 +158,9 @@ def load_genes(self, cursor):
                         new_articles += 1
                     gene.articles.add(article)
 
-                sys.stdout.write('scanned: {0}\tgenes: {1}\tsynonyms: {2}\tarticles: {3}\r'.format(count, new_genes, new_synonyms, new_articles))
-                sys.stdout.flush()
+                log.debug(
+                    'scanned: {0}\tgenes: {1}\tsynonyms: {2}\tarticles: {3}\r'
```
bruth: I am not sure if the carriage return should be here now that it's not guaranteed to do to stdout. If the logger adds a newline after each statement (which I presume it does), the carriage return will have no effect.

varify/management/commands/subcommander.py
```diff
@@ -25,24 +28,29 @@ def usage(self, subcommand):
 
     def print_help(self, prog_name, subcommand):
         super(Subcommander, self).print_help(prog_name, subcommand)
-        sys.stdout.write('{0}\n\n'.format(self.print_subcommands(prog_name)))
+        log.debug('{0}\n\n'.format(self.print_subcommands(prog_name)))
```
bruth: This one should not be changed since this is a base class for Django commands.

varify/management/commands/sqldropindexes.py
```diff
@@ -5,6 +6,8 @@
 from django.core.management.base import AppCommand
 from django.core.management.color import no_style
 
+log = logging.getLogger(__name__)
```
bruth: The logger is not being used anywhere in this module. Also this is a Django command, so output should be to stdout.

bruth: Ignore the second sentence.. it should probably only be to stdout for non-long running commands.

varify/phenotypes/management/subcommands/load.py
```diff
@@ -103,8 +101,7 @@ def load_hgmd_phenotypes(label, keys, cursor, using):
                 if count % BATCH_SIZE == 0:
                     transaction.commit()
 
-        sys.stdout.write('Loading {0}...{1}/{2}\r'.format(label, count, total))
-        sys.stdout.flush()
+        log.debug('Loading {0}...{1}/{2}\r'.format(label, count, total))
```
bruth: Another carriage return to remove.

varify/phenotypes/management/subcommands/load.py
```diff
@@ -332,6 +329,6 @@ def handle(self, *args, **options):
                     handler(cursor, using)
                     transaction.commit()
                 except Exception, e:
-                    print 'Failed to load {0}'.format(handler.short_name)
+                    log.error('Failed to load {0}'.format(handler.short_name))
                     transaction.rollback()
                     log.exception(e)
```
bruth: `log.error` can be remove and the message passed in this call since `log.exception` already includes the traceback. Having the two calls is redundant.

varify/raw/management/subcommands/load.py
```diff
@@ -33,56 +36,55 @@ class LoadCommand(BaseCommand):
...
             files, options = targets[target]
             try:
                 self.load_files(cursor, target, files, options)
-                print('- {0}'.format(target))
+                log.debug('- {0}'.format(target))
             except Exception:
                 traceback.print_exc()
-                print('----')
-                print('Error loading {0}, rolling back'.format(target))
+                log.error('Error loading {0}, rolling back'.format(target))
```
bruth: Instead of printing the traceback, a `log.exception` should be called.

varify/raw/management/subcommands/load.py
```diff
@@ -33,56 +36,55 @@ class LoadCommand(BaseCommand):
...
         for target in targets:
             try:
                 self.create_table(cursor, target, targets[target][1])
-                print('- {0}'.format(target))
+                log.debug('- {0}'.format(target))
             except Exception:
                 traceback.print_exc()
-                print('----')
-                print('Error creating {0}, rolling back'.format(target))
+                log.error('Error creating {0}, rolling back'.format(target))
```
bruth: Instead of printing the traceback, `log.exception` should be called.

varify/raw/management/subcommands/load.py
```diff
@@ -33,56 +36,55 @@ class LoadCommand(BaseCommand):
...
         for target in targets:
             try:
                 self.drop_table(cursor, target, targets[target][1])
-                print('- {0}'.format(target))
+                log.debug('- {0}'.format(target))
             except Exception:
                 traceback.print_exc()
-                print('----')
-                print('Error dropping {0}, rolling back..'.format(target))
+                log.error('Error dropping {0}, rolling back..'.format(target))
```
bruth: Same as the below two.. use `log.exception`.

varify/samples/management/subcommands/prioritize.py
```diff
@@ -41,6 +42,6 @@ def handle(self, *args, **options):
             count += 1
 
         if count:
-            print '{0} variants prioritized\n'.format(count)
+            log.debug('{0} variants prioritized\n'.format(count))
```
bruth: Newline isn't needed.

PR #79, opened 2014-02-10
Make assessment.assessment_category a required column

### LINKED ISSUE #76: Allow for category selection for all pathogenicities
Currently, there is a subset of pathogenicities that make the category selection disabled. The category should become a required field and be enabled for all pathogenicity values. We will need to add a category called 'Other' to account for the fact that it is now a required field. This should be the default value for the migration that will need to happen on the assessments table if the category is null now.

naegelyd:
Fix #76.

The addition of the "Other" assessment(via a data migration in this change set) will account for any existing assessments that have a null assessment_category field. This change will allow the UI to make category a selectable field for all pathogenicities instead of the limited subset as it is now.


varify/assessments/migrations/0011_add_other_assessment_category.py
```diff
@@ -0,0 +1,244 @@
...
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+    no_dry_run = True
+
+    def forwards(self, orm):
+        AssessmentCategory = orm['assessments.assessmentcategory']
+        if not AssessmentCategory.objects.filter(pk=7).exists():
```
bruth, line 12: Is there any reason you are filtering by primary key? It would be safer to filter on the `name` which is required to be unique anyway.

varify/assessments/migrations/0012_auto__chg_field_assessment_assessment_category.py
```diff
@@ -0,0 +1,245 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
+        # Changing field 'Assessment.assessment_category'
+        db.alter_column('assessment', 'assessment_category_id', self.gf('django.db.models.fields.related.ForeignKey')(default=lambda: orm['assessments.AssessmentCategory'].objects.get(id=7), to=orm['assessments.AssessmentCategory']))
```
bruth, line 13: Same as above, filtering by name would be safer.

varify/assessments/migrations/0011_add_other_assessment_category.py
naegelyd, line 12: The UI assumes a fixed ID(maybe it shouldn't either) so I just used the ID here too as they should not deviate from what we expect and they are all created with fixed IDs. I can change these to use names instead but no one should be adding categories without a migration and update to the UI so I figured it was "safe" to just use the ID here.

PR #112, opened 2014-02-19
Force user to select a sample

### LINKED ISSUE #64: Require sample to be selected
There should not be any calls to the `preview` endpoint on the server unless the user has selected a sample. Eventually this might be a config option in Cilantro but until that is fully implemented and thought out, we can just block the calls in Varify if no sample is selected in the user's current filter set.

naegelyd:
Fix #64.

This has the side effect of blocking all context changes until the sample is selected but that is necessary given the way required concepts are implemented currently in Cilantro. There is a notification made to the user when they try to modify the context with no sample selected so it should be clear enough that no one is confused.


varify/static/scripts/coffeescript/main.coffee
```diff
@@ -62,6 +62,18 @@ require
     c.config.set('fields.instances.58.form.controls', ['Sift'])
     c.config.set('fields.instances.56.form.controls', ['PolyPhen'])
 
+    # Mark the Sample concept as required and display a notification to the
+    # user when it is not populated.
+    c.config.set('query.concepts.required', [2])
+    c.on c.CONTEXT_INVALID, (concepts) ->
```
bruth: There is a `c.CONTEXT_REQUIRED` event as well when the sample is attempted to be removed.

Issue #25, opened 2014-01-13
Add optimize dependency to build target in Makefile


naegelyd:
The optimize dependency needs to be added to the build target [here](https://github.com/cbmi/varify/blob/harvest/Makefile#L26-L27). This should probably be done after #24.

naegelyd:
Fixed in #129. It's mostly a dupe of #59 anyway.

PR #129, opened 2014-02-25
Add minified JS and step to create in on build target in Makefile

### LINKED ISSUE #59: Add minified scripts to the repo
If these aren't present and varify is used in an environment where `DEBUG=False` then some scripts will not be found.

naegelyd:
Fix #59, #25.


varify/static/scripts/javascript/app.build.js
```diff
@@ -0,0 +1,23 @@
...
+    config: {
+        tpl: {
+            variable: 'data'
+        }
+    },
+
+    dir: 'min',
+
+    modules: []
+})
```
bruth, line 31: Usually when doing cross-project dependencies, I've needed to add the `paths` option with the items like `cilantro: 'empty:'` to prevent external dependencies from being minified in.

Issue #132, opened 2014-02-26
Genotype not being set when loading results


naegelyd:
The samples and results load as expected but the genotype field on the results is null for all results.

mitalia:
This may also include coverage

PR #136, opened 2014-02-27
added unusual genotyping calls to fixture

leipzig:
fix for https://github.com/cbmi/varify/issues/132


naegelyd:
Unfortunately, this is just a start to fixing #132. The call data is not being read correctly. I already integrated these fixture changes into an upcoming PR addressing #132 so I am closing this in favor of the larger fix PR which will be coming soon.

PR #139, opened 2014-03-04
Add command and script for updating gene rankings

### LINKED ISSUE #41: Add phenotype check to loaders
When loaders for new samples fire, we should check to see if phenotypes are available. If they are, we should automatically assign scores to variants using the variant scoring service.

naegelyd:
Fix #41.

This does not directly hook in to the loader. Instead, this can be run as a cron job like `recalculate_frequencies.py` is run now. There is no guarantee that anyone has setup phenotype and gene ranking endpoints so we do not automatically run this during the pipeline loading process.

### DIFF
--- a/varify/samples/management/subcommands/gene_ranks.py
+++ b/varify/samples/management/subcommands/gene_ranks.py
@@ -222,6 +222,7 @@ class Command(BaseCommand):
                         log.exception("Error saving gene ranks and scores for "
                                       "sample '{0}'".format(sample.label))
                         transaction.rollback()
+                        continue
 
                     transaction.commit()
 
--- a/varify/samples/management/subcommands/gene_ranks.py
+++ b/varify/samples/management/subcommands/gene_ranks.py
@@ -1,0 +1,213 @@
+import json
+import logging
+import requests
+from datetime import datetime
+from django.conf import settings
+from optparse import make_option
+from django.db import transaction, DEFAULT_DB_ALIAS
+from django.core.management.base import BaseCommand
+from requests.exceptions import SSLError, ConnectionError, RequestException
+from varify.samples.models import Sample
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    option_list = BaseCommand.option_list + (
+        make_option('--database', action='store', dest='database',
+                    default=DEFAULT_DB_ALIAS,
+                    help='Specifies the target database to load results.'),
+        make_option('--force', action='store_true', dest='force',
+                    default=False,
+                    help='Forces recomputation of all gene rankings')
+    )
+
+    def handle(self, **options):
+        if not settings.PHENOTYPE_ENDPOINT:
+            log.error('PHENOTYPE_ENDPOINT must be defined in settings for '
+                      'gene rankings to be updated.')
+            return
+
+        if not settings.GENE_RANK_BASE_URL:
+            log.error('GENE_RANK_BASE_URL must be defined in settings for '
+                      'gene rankings to be updated.')
+            return
+
+        if not settings.VARIFY_CERT or not settings.VARIFY_KEY:
+            log.error('VARIFY_CERT and VARIFY_KEY must be defined in settings '
+                      'for gene rankings to be updated.')
+            return
+
+        database = options.get('database')
+        force = options.get('force')
+
+        # Construct the cert from the setting to use in requests to the
+        # phenotype endpoint.
+        cert = (settings.VARIFY_CERT, settings.VARIFY_KEY)
+
+        # We ignore all the samples that aren't published. They aren't visible
+        # to the user so we don't bother updating related scores.
+        samples = Sample.objects.filter(published=True)
+
+        updated_samples = 0
+        total_samples = 0
+        for sample in samples:
+            total_samples += 1
+
+            # Construct the URL from the setting and the sample label. The
+            # sample label is used to retrieve the phenotype info on the remote
+            # endpoint.
+            url = settings.PHENOTYPE_ENDPOINT % sample.label
+
+            # Get the phenotype information for this sample. If the
+            # phenotype is unavailable then we can skip this sample.
+            try:
+                response = requests.get(url, cert=cert, verify=False)
+            except SSLError:
+                log.exception('Skipping sample "{0}". An SSLError occurred '
+                              'during phenotype retrieval request.'
+                              .format(sample.label))
+                continue
+            except ConnectionError:
+                log.exception('Skipping sample "{0}". A ConnectionError '
+                              'occurred during phenotype retrieval request.'
+                              .format(sample.label))
+                continue
+            except RequestException:
+                log.exception('Skipping sample "{0}". The sample has no '
+                              'phenotype data associated with it'
+                              .format(sample.label))
+                continue
+
+            try:
+                phenotype_data = json.loads(response.content)
+            except ValueError:
+                log.error("Could not parse response from {0}, skipping '{1}'."
+                          .format(url, sample.label))
+                continue
+
+            try:
+                phenotype_modified = datetime.strptime(
+                    phenotype_data['last_modified'], "%Y-%m-%dT%H:%M:%S.%f")
+            except ValueError:
+                phenotype_modified = datetime.min
+                log.warn("Could not parse 'last_modified' field on phenotype "
+                         "data. Using datetime.min so that only unranked "
+                         "samples will be ranked. If the 'force' flag was "
+                         "used then all samples will be updated despite this "
+                         "parsing failure.")
+
+            if (not force and sample.phenotype_modified and
+                    sample.phenotype_modified < phenotype_modified):
+                log.debug("Sample '{0}' is already up to date, skipping it."
+                          .format(sample.label))
+                continue
+
+            # Extract the HPO terms from the data returned from the phenotype
+            # endpoint. We need to modify the terms a bit because the phenotype
+            # endpoint has terms in the form 'HP_0011263' and the gene ranking
+            # enpoint expects them to be of the form 'HP:0011263'.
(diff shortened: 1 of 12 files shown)


varify/samples/receivers.py
```diff
@@ -16,14 +16,20 @@
...
         world_cohort = Cohort.objects.get(**lookup)
     except Cohort.DoesNotExist:
-        world_cohort = Cohort(**lookup)
-        world_cohort.save()
+        if getattr(settings, 'VARIFY_AUTO_CREATE_COHORT', True):
+            world_cohort = Cohort(**lookup)
+            world_cohort.save()
+        else:
+            log.info("World cohort was not found and was not created because "
+                     "VARIFY_AUTO_CREATE_WORLD_COHORT setting is False.")
```
bruth: Where is this setting applied? Did you mean `VARIFY_AUTO_CREATE_COHORT`?

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+        # were sample labels supplied as arguments then we limit the rankings
+        # updates to those samples, otherwise we process all samples.
+        samples = Sample.objects.filter(published=True)
+
+        if args:
+            samples = samples.filter(label__in=args)
+
+        updated_samples = 0
+        total_samples = 0
+        for sample in samples:
```
bruth, line 62: Empty line above for clarity.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+        make_option('--database', action='store', dest='database',
+                    default=DEFAULT_DB_ALIAS,
+                    help='Specifies the target database to load results.'),
+        make_option('--force', action='store_true', dest='force',
+                    default=False,
+                    help='Forces recomputation of all gene rankings')
+    )
+
+    def handle(self, *args, **options):
+        if not settings.PHENOTYPE_ENDPOINT:
```
bruth: Unless they settings are defined in the `global_settings.py` for Varify, doing this will cause an `AttributeError`.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+
+        updated_samples = 0
+        total_samples = 0
+        for sample in samples:
+            total_samples += 1
+
+            # Construct the URL from the setting and the sample label. The
+            # sample label is used to retrieve the phenotype info on the remote
+            # endpoint.
+            url = settings.PHENOTYPE_ENDPOINT % sample.label
```
bruth: The new-style string formatting `settings.PHENOTYPE_ENDPOINT.format(sample.label)` is preferred.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+            try:
+                phenotype_data = json.loads(response.content)
+            except ValueError:
+                log.error("Could not parse response from {0}, skipping '{1}'."
+                          .format(url, sample.label))
+                continue
+
+            try:
+                phenotype_modified = datetime.strptime(
+                    phenotype_data['last_modified'], "%Y-%m-%dT%H:%M:%S.%f")
```
bruth, line 99: Is there control over the format of this datetime string (obviously at the phenotype endpoint)? Ideally if would be in the standard ISO 8601 format.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                phenotype_modified = datetime.min
+                log.warn("Could not parse 'last_modified' field on phenotype "
+                         "data. Using datetime.min so that only unranked "
+                         "samples will be ranked. If the 'force' flag was "
+                         "used then all samples will be updated despite this "
+                         "parsing failure.")
+
+            # If the parsed response doesn't contain any HPO terms then we can
+            # skip this sample since we cannot rank genes without HPO terms.
+            if not phenotype_data.get('hpoAnnotations', []):
```
bruth: `data.get('...')` returns `None` as the default value, so there is no need to use a list as the default value.

varify/samples/management/subcommands/gene_ranks.py
naegelyd, line 99: @jeffmax Can we change the datetime string on the phenotype endpoint to ISO 8601 as suggested above? This code here is built to parse the datetimes I am getting out of the phenotype endpoint as it stands now.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                continue
+
+            # Extract the HPO terms from the data returned from the phenotype
+            # endpoint. We need to modify the terms a bit because the phenotype
+            # endpoint has terms in the form 'HP_0011263' and the gene ranking
+            # enpoint expects them to be of the form 'HP:0011263'.
+            hpo_terms = []
+            for hpo_annotation in phenotype_data['hpoAnnotations']:
+                try:
+                    hpo_id = str(hpo_annotation.get('hpo_id', ''))
```
bruth: Is there a reason for the `str` call? Would this be a string no matter what?

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                log.exception('Error retrieving gene rankings, skipping '
+                              'sample "{0}".'.format(sample.label))
+                continue
+
+            gene_data = json.loads(gene_response.content)
+            ranked_genes = gene_data['ranked_genes']
+
+            updated_results = 0
+            total_results = 0
+            for result in sample.results.all():
```
bruth, line 173: Line above for readability.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+
+                with transaction.commit_manually(database):
+                    try:
+                        # Get the gene for this result. Since a result can
+                        # have more than one gene associated with it, we
+                        # return the first gene symbol in the list. This is
+                        # the same one that will be shown in the collapsed
+                        # gene list on the variant row in the results table.
+                        gene = result.variant.effects.values_list(
+                            'transcript__gene__symbol', flat=True)[0]
```
bruth: This will cause an index error if not genes are found which prevents the next condition `if not gene` from logging the correct message.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                        if not gene:
+                            log.debug("Result with id {0} has no gene, "
+                                      "skipping result.".format(result.id))
+                            transaction.rollback()
+                            continue
+
+                        # Get the first item in the ranked gene list with a
+                        # symbol matching the gene we looked up above for this
+                        # result.
+                        ranked_gene = next(
```
bruth: This is a tad convoluted and would cause a `StopIteration` error if no gene matches. Why not:

``` python
ranked_gene = None

for ranked_gene in ranked_genes:
    if ranked_gene.get('symbol', '').lower() == gene.lower():
        break
else:
    log.debug('...')
    # etc.
```

bruth: Modified the code block above to bind to `ranked_gene` for downstream use.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                        if not ranked_gene:
+                            log.debug("Could not find '{0}' in ranked gene "
+                                      "list, skipping result".format(gene))
+                            transaction.rollback()
+                            continue
+
+                        try:
+                            rs = ResultScore.objects.get(result=result)
+                            rs.rank = ranked_gene.get('rank', None)
+                            rs.score = ranked_gene.get('score', None)
```
bruth: No need to the above two default `None` values.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                        except ResultScore.DoesNotExist:
+                            rs = ResultScore(
+                                result=result,
+                                rank=ranked_gene.get('rank', None),
+                                score=ranked_gene.get('score', None))
```
bruth: Same as above.

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+                        rs.save()
+
+                        updated_results += 1
+
+                    except Exception:
+                        log.exception("Error saving gene ranks and scores for "
+                                      "sample '{0}'".format(sample.label))
+                        transaction.rollback()
```
bruth, line 235: I believe there is a missing `continue` statement here after this rollback. Otherwise a commit occurs below..

varify/samples/management/subcommands/gene_ranks.py
```diff
@@ -0,0 +1,235 @@
...
+
+                    transaction.commit()
+
+            sample.phenotype_modified = datetime.now()
```
bruth, line 240: Should this only be updated if at least one result has been updated?

varify/samples/receivers.py
naegelyd: I did. Guess my indecision over the setting name is now evident. Good catch.

varify/samples/management/subcommands/gene_ranks.py
naegelyd, line 240: The reason this is updated is to account for the case where a sample has results but no ranked genes(that is, the ranking endpoint considers all the results' genes invalid for the given HPO terms). In this case, the sample is updated in that none of the results were ranked. While I doubt this would be very common, it is possible.

PR #142, opened 2014-03-05
Move templates from cilantro static into varify proper

### LINKED ISSUE #24: Move templates from Cilantro folder to Varify folder
This was originally recommended [here](https://github.com/cbmi/varify/pull/23#issuecomment-32027586) but it was decided to do the move in a separate branch(thus this ticket) and not part of the custom review workflow PR. Moving the templates will allow for easy transitions to new Cilantro versions because we will not need to worry about maintaining any Varify specific files in the Cilantro directory.

naegelyd:
Fix #24.


varify/static/scripts/javascript/src/templates/workflows/results.html
```diff
@@ -0,0 +1,150 @@
+<div class=row-fluid>
+    <div class="span12 hide serrano-version-warning alert">
```
bruth: This should no longer be necessary with the new Cilantro notifications.

PR #155, opened 2014-03-09
Vcf export

awenocur:
Partially functioning VCF exporter: generates a VCF file with a row for each selected variant

As of this request, the vcf_export branch is lacking some important functionality, however should work well enough for us to start testing the exporter on the dev instance.


naegelyd:
Haven't reviewed the code changes yet but thesse changes will need to be updated to be flake8 compliant. That is why the build is failing. Also, the exporter should have a test added for it.

awenocur:
I don't see any examples of tests for exporters in avocado or chop-varify.  Would this be the first?  If so, where would it go?

naegelyd:
Avocado tests its own exporters as seen [here](https://github.com/cbmi/avocado/blob/master/tests/cases/exporting/tests.py). The tests for varify's exporters should go in [this folder](https://github.com/cbmi/varify/tree/master/tests/cases) under a new `export` folder.

naegelyd:
The original comment says: 

`As of this request, the vcf_export branch is lacking some important functionality...`. 

What functionality is missing? Tests aside, I am hesitant to merge this if features are missing. Feature branches like this should be completed(on the branch itself) before being merged into the stable master branch.

awenocur:
will reopen when mature

awenocur:
added package to incorporate test case when ready

PR #156, opened 2014-03-10
Require one or more columns to be selected before exporting

### LINKED ISSUE #144: Require user to select at least one column when exporting
Don't make the call to the server if the user has no columns selected(that is, their view is empty). This will just result in empty files so the user should be forced to select at least one column first.

naegelyd:
Fix #144.


varify/static/scripts/coffeescript/ui/workflows/results.coffee
```diff
@@ -163,9 +163,15 @@ define [
             , 25
 
         onExportClicked: =>
+            # If there are no columns selected then we should not try to
+            # export since the user will just get an empty file or empty
+            # template. Show an error with instructions on adding a column.
+            if @columns.currentView.data.facets.length == 0
+                @$('#export-error-message').html('One or more columns must be selected. Open the &quot;Columns&quot; tab, add at least one column using the green &quot;plus&quot; buttons next to the column names, and click &quot;Export&quot; to try again.')
```
bruth: Should this say "Click" instead of "Open" since the modal is already open?

bruth:
Is there a condition to hide the error message once a column is selected? This could be done in the columns tab directly or by observing the `facets` collection.

PR #175, opened 2014-03-17
Port utils and models to JS

naegelyd:


varify/static/scripts/javascript/src/models.js
```diff
@@ -1,8 +1,14 @@
...
+    'underscore',
+    './models/result',
+    './models/assessment',
+    './models/phenotype'
+], function(_) {
+
+    // Modules to be mixed-in with exports
+    var mods = Array.prototype.slice.call(arguments, 1);
+
+    return $.extend.apply(null, mods);
```
bruth: `$` is being accessed as a global. I think you meant to do:

``` javascript
    // Get the slice of arguments that represent the modules
    // (after underscore)
    var mods = [].slice.call(arguments, 1);

    // Merge the mods into an empty object that will be exported
    return _.extend.apply(_, [{}].concat(mods));
```

varify/static/scripts/javascript/src/models/assessment.js
```diff
@@ -1,74 +1,67 @@
...
+                    'assessment_category',
+                    'father_result',
+                    'mother_result',
+                    'pathogenicity'
+                ];
 
-    return Assessment;
+                _.each(objFields, function(field) {
+                    if (response[field]) {
+                        return data[field] = response[field]['id'];
```
bruth: No need for the `return` on the assignment. Also, using attribute-based access is generally preferred over key-based: `response[field].id`.

varify/static/scripts/javascript/src/models/result.js
```diff
@@ -1,51 +1,50 @@
...
+            variant.effects = _.sortBy(variant.effects, function(eff) {
+                return utils.effectImpactPriority(eff.impact);
+            });
+
+            var uniqueGenes = {},
+                genes = [];
 
-    return Result;
+            _.each(variant.effects, function(eff) {
+                if (!eff.transcript || !(gene = eff.transcript.gene)) {
```
bruth: This is a bit of a preference thing, but _early exit_ returns look nice on a single line.

``` javascript
if (!eff.transcript || !(gene = eff.transcript.gene)) return;
```

varify/static/scripts/javascript/src/models/result.js
```diff
@@ -1,51 +1,50 @@
...
-  })(Backbone.Model);
-  return {
-    Result: Result
-  };
+                if (/^LOC\d+/.test(gene.symbol) || (uniqueGenes[gene.symbol] != null)) {
+                    return;
+                }
+
+                uniqueGenes[gene.symbol] = true;
+                return genes.push({
```
bruth, line 31: Not necessarily better, but an alternative, `_.pick(gene, 'symbol', 'hgnc_id', 'name')`

varify/static/scripts/javascript/src/models.js
naegelyd: Good catch.

varify/static/scripts/javascript/src/models/result.js
naegelyd, line 31: I'm going to leave this as is for clarity. If you feel strongly in favor of `pick` we can switch it but for now I will leave as is.

PR #185, opened 2014-03-25
Add link for viewing pedigree PDF

naegelyd:


varify/samples/resources.py
```diff
@@ -292,12 +292,43 @@ def get(self, request, sample_id):
         return response.content
 
 
+class PedigreeResource(ThrottledResource):
+    def get(self, request, year, month, day, f):
```
bruth: What is `f`? Format? Either a comment or a more description name should be used.

PR #186, opened 2014-03-25
Integrate SolveBio Clinvar into variant details (and condensed view)

davecap:
Adds a requirement for the SolveBio Python client, which contains a small wrapper for Django apps. The Django module allows SolveBio datasets to be locally aliased, so that a query to "clinvar" will route to "clinvar/0.0.2/clinvar" in SolveBio. The query is integrated into the variant API resource and filters clinvar by chromosomal position, gene symbol, and HGVS notation.

Dataset aliases can be set in a Django setting (`SOLVEBIO_DATASETS`) and via the Django Admin (see below). The Django setting overrides the values in the database.

screen shot 2014-03-25 at 3 24 08 pm


naegelyd:
Haven't started looking at this yet but I wanted to let you know that I am one file away from being fully moved from coffeescript to javascript so this will need to be rebased against master once that happens and all coffeescript moved over to JS. I expect to have that work done tomorrow so you should be able to rebase then.

naegelyd:
Also, the commits will need to be amended with [signoffs](https://github.com/cbmi/varify/blob/master/CONTRIBUTING.md#contributing) in order for the build to run and to be considered for merging. Looking forward to getting this code into Varify. Thanks again for the PR.

davecap:
Sounds good, I'll re-submit a new pull request once the JS work is finished.

PR #192, opened 2014-03-28
Move all coffeescript to JS

### LINKED ISSUE #143: Port all coffeescript to javascript
Cilantro is slowly undergoing this change. There is no immediate rush on this but since Cilantro is doing it, Varify might as well follow along and move to JS as well.

naegelyd:
Fix #143.

Also, this updates the Cilantro version and updates Varify code to match changes in Cilantro.


varify/static/scripts/javascript/src/models/phenotype.js
```diff
@@ -6,9 +6,63 @@ define([
...
             return utils.toAbsolutePath(path);
+        },
+
+        parsePriority: function(value) {
+            return parseInt(value.priority, 10) || this.lowestPriority + 1;
+        },
+
+        parse: function(attrs) {
+            var path, lastModified, phenotypeModified;
+            //Backbone.Model.prototype.parse.call(this, attrs);
```
bruth: You can remove this line since `Backbone.parse` is a no-op.

varify/static/scripts/javascript/src/models/phenotype.js
```diff
@@ -6,9 +6,63 @@ define([
...
+        lowestPriority: 10,
+        highestPriority: 1,
+
         urlRoot: function() {
             var path = 'api/samples/' + this.get('sample_id') + '/phenotypes/';
             return utils.toAbsolutePath(path);
+        },
+
+        parsePriority: function(value) {
+            return parseInt(value.priority, 10) || this.lowestPriority + 1;
```
bruth, line 19: I realize this may be domain-specific, but why does this default to lowest plus one? Maybe a comment would help.

naegelyd, line 19: Ha, I wish I knew the answer to this. I was just copying over code that @jeffmax wrote. @jeffmax can you drop a note here explaining the `+ 1` here and I will add the comment to the code.

varify/static/scripts/javascript/src/ui/controls.js
```diff
@@ -1,8 +1,13 @@
-// Generated by CoffeeScript 1.7.1
-var __slice = [].slice;
+/* global define */
+
+define([
+    'underscore',
+    './controls/sift',
+    './controls/polyphen'
+], function() {
```
bruth: `_` must be the first argument here to prevent use of a global.

varify/static/scripts/javascript/src/ui/exporter.js
```diff
@@ -0,0 +1,33 @@
+/* global define */
+
+define([
+    'cilantro'
+], function(c) {
+
+    var ExporterDialog = c.ui.ExporterDialog.extend({
+        template: 'varify/export/dialog',
+
+        _events: {
```
bruth: The `events` hash can be defined as a function since Backbone uses [`_.result`](http://underscorejs.org/#result) to resolve the object:

``` javascript
events: function() {
    return _.extend({ ... }, c.ui.ExporterDialog.prototype.events);
}
```

varify/static/scripts/javascript/src/ui/exporter.js
```diff
@@ -0,0 +1,33 @@
...
+            'click [data-action=change-columns]': 'changeColumnsClicked'
+        },
+
+        initialize: function() {
+            this.events = _.extend({}, this._events, this.events);
+
+            c.ui.ExporterDialog.prototype.initialize.call(this);
+        },
+
+        onRender: function() {
```
bruth: This doesn't extend anything, so it shouldn't be necessary.

varify/static/scripts/javascript/src/ui/modals.js
```diff
@@ -1,8 +1,13 @@
-// Generated by CoffeeScript 1.7.1
-var __slice = [].slice;
+/* global define */
+
+define([
+    'underscore',
+    './modals/result',
+    './modals/phenotype'
+], function() {
```
bruth: Define `_` as an argument.

varify/static/scripts/javascript/src/ui/modals/phenotype.js
```diff
@@ -0,0 +1,291 @@
...
+        },
+
+        retrievePhenotypes: function(recalculateRankings) {
+            if (recalculateRankings == null) {
+                recalculateRankins = false;
+            }
+
+            // We are about to attempt to reload the phenotypes so we want to
+            // start fresh and clear any old artifacts.
+            this.ui.warning.hide();
```
bruth: Would it be worth having a class applied to this view that hides all of these elements?

varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -1,683 +1,859 @@
...
+/* global define */
+
+define([
+    'jquery',
+    'underscore',
+    'marionette',
+    '../../models',
+    '../../utils',
+    '../../templates',
+    'cilantro/utils/numbers'
```
bruth: Using the minified Cilantro build, this will load all of Cilantro anyway. All the number utilities are accessible at `c.utils`.

varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -1,683 +1,859 @@
...
-        valid = false;
-        this.errorMsg.append('<h5>Please select one of the &quot;Sanger Requested&quot; options.</h5>');
-      }
-      if (!valid) {
-        this.errorContainer.show();
-      }
-      return valid;
-    };
+                        if (eff.hgvs_c) {
+                            content.push('' + eff.hgvs_c + ' ');
```
bruth, line 292: I've seen this in various places. I don't believe (but since since it's JS I wouldn't be surprised) a prefix string concatenation does anything different from a postfix one. Since you concat a space here, the former should not be needed.

bruth:
This is out of the scope of this PR, but there are many places where HTML is being concatenated and inserted. These views should be changed to layouts and the larger HTML strings change to their own views. This enables using templates (and therefore overriding them using the Cilantro template API by projects) and handles a lot of the `if (foo && foo.length)` handling since collection views handle all that already.

naegelyd:
Agree with HTML concatenation issue addressed above. In this PR, I just wanted to get this code ported over to get off of Coffeescript. Going forward, there are a lot of opportunities for improving the JS especially those you mention above.

varify/static/scripts/javascript/src/ui/modals/result.js
naegelyd, line 292: In this case, we could probably get away without the leading `''` but in some cases it is essential. For example, this example that I copied from my console. The postfix, in this case, just turns the leading result into a string while the prefix uses string concatenation throughout.

``` javascript
'' + 1 + 2 + 3
"123"
1 + 2 + 3 + ''
"6"
```

varify/static/scripts/javascript/src/ui/tables.js
```diff
@@ -1,8 +1,15 @@
...
-var __slice = [].slice;
+/* global define */
+
+define([
+    'underscore',
+    './tables/body',
+    './tables/header',
+    './tables/row',
+    './tables/table'
+], function() {
```
bruth: `_` missing.

varify/static/scripts/javascript/src/ui/tables/header.js
```diff
@@ -1,107 +1,121 @@
...
-        if (f.get('concept') === concept) {
-          direction = f.get('sort');
-          if (direction != null) {
-            if (direction.toLowerCase() === "asc") {
-              f.set('sort', "desc");
-              return f.set('sort_index', 0);
-            } else {
-              f.unset('sort');
-              return f.unset('sort_index');
+            if ((concept != null) && !isNaN(concept)) {
```
bruth, line 28: Should be `_.isNaN`?

bruth, line 28: Nevermind. Native function.. I'm a noob.

varify/static/scripts/javascript/src/ui/workflows/results.js
```diff
@@ -1,684 +1,101 @@
...
-        this.changeExportStatus(exportTypeTitle, "timeout");
-      } else {
-        this.changeExportStatus(exportTypeTitle, "success");
-      }
-      if (this.numPendingDownloads === 0) {
-        this.$('[data-toggle=export-options]').prop('disabled', false);
-        return this.$('.export-progress-container').hide();
-      }
-    };
+        _events: {
```
bruth: Same as above regarding using a function.

varify/static/scripts/javascript/src/ui/workflows/results.js
```diff
@@ -1,684 +1,101 @@
...
-        _results = [];
-        for (i = _i = 0, _ref = selectedTypes.length - 1; _i <= _ref; i = _i += 1) {
-          this.changeExportStatus(this.$(selectedTypes[i]).attr('title'), "pending");
-          _results.push(setTimeout(this.startExport, i * delay, selectedTypes[i], pagesSuffix));
-        }
-        return _results;
-      }
-    };
+        onRender: function() {
+            $(document).on('scroll', this.onPageScroll);
```
bruth, line 72: This should not matter in practice since workflows are not closed.. but this handle should be unbound `onClose`.

naegelyd, line 72: Won't this be unbound in [c.ui.ResultWorkflow.onClose](https://github.com/cbmi/cilantro/blob/master/src/js/cilantro/ui/workflows/results.js#L238-L240) since I am extending it?

varify/static/scripts/javascript/src/utils.js
```diff
@@ -2,6 +2,54 @@
...
+    };
+
+
+    var getRootUrl = function() {
+        /*
+         * Get the route-free URL. That is, we want to remove the route at the
+         * end of the URL and be left with the root URL so we can use this to
+         * construct the result URLs later on.
+         */
+        return window.location.href.replace(new RegExp('/[^/]*/$'), '/');
```
bruth: Literals are preferred over use of the constructor: `/[^\/]*\/$/`, but I get the desire not to have to escape the forward slashes.

varify/static/scripts/javascript/src/utils.js
```diff
@@ -168,13 +168,44 @@ define([], function() {
...
+         * Utility method for retrieving the full list of sample labels in
+         * the current context. Given the current sample control, this could
+         * be 0, 1, or more sample labels. This method returns a list of the
+         * sample labels in the supplied context. If there is an issue
+         * retreving them or there are no sample labels, an empty list is
+         * returned.
+         */
+        var samples = [], json;
+
+        if (context  && (json = context.get('json'))) {
```
bruth: Extra space after `context`.

varify/static/scripts/javascript/src/ui/modals/result.js
bruth, line 292: Ah right. Most of this should all go away when moving to a template-centric approach. I wouldn't worry about this now.

varify/static/scripts/javascript/src/ui/workflows/results.js
bruth, line 72: Yes, you are right. Forgot about that.

varify/static/scripts/javascript/src/models/phenotype.js
jeffmax, line 19: Sorry, I am just discovering I missed all these notifies. I think that +1 is a ill-conceived idea I had to make sure that values that were not integers came below everything else, but it's highly unlikely to happen and confusing, and even if it does happen, the lowest priority is fine. Will remove it.

PR #193, opened 2014-03-28
integrate SolveBio ClinVar into the collapsed result set and variant sum...

davecap:
Adds a requirement for the SolveBio Python client, which contains a small wrapper for Django apps. The Django module allows SolveBio datasets to be locally aliased to avoid hard-coding dataset versions into the code. The ClinVar query is integrated into the variant API resource and filters ClinVar by chromosomal position, gene symbol, and HGVS notation.

Dataset aliases can be set in a Django setting (`SOLVEBIO_DATASETS`) and via the Django Admin (see below). The Django setting overrides the values in the database.


varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -512,7 +545,7 @@ define([
 
             $row1.append(this._span(this.renderSummary(this.model.attributes, attrs), 3));
             $row1.append(this._span(this.renderEffects(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
-            $row1.append(this._span(this.renderPhenotypes(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
+            $row1.append(this._span(this.renderPhenotypes(attrs) + this.renderClinvar(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
```
naegelyd: Let's move this to `$row2`. The layout of the details tab has changed a bit recently and there is an empty space at the end of "row2" that we made in anticipation of this change. From the picture before you can see that it will go to the right of the `Articles` section. Also, if you think that this section will be tall(height > 300px) you can take advantage of the built-in expand/collapse functionality by appending `.addClass('expandable-details-item').append(this._renderExpandCollapse));` to the line where you are adding the clinvar render call to row2.

screen shot 2014-03-28 at 3 45 47 pm

varify/variants/resources.py
```diff
@@ -69,6 +77,21 @@ def get(self, request, pk):
 
         data['cohorts'] = cohort_list
 
+        # SolveBio integration
```
naegelyd: If `SOLVEBIO_API_KEY` is `None`, then this code should not execute. I have your PR running locally and don't have the api key set. When the key is not set, the results table doesn't render because each of the `api/samples/variants/<id>/` calls fail with the following error. While we would expect this error since the key isn't set, it would be better if there is no call made to solvevio resources when the `SOLVEBIO_API_KEY` setting is not valid.

```
Sorry, your API credentials seem to be invalid.

SolveBio is currently in Private Beta.
Please go to https://www.solvebio.com to find out more.

If you are a beta user, please log in by typing:

    solvebio login
```

davecap: Good point I'll add some error checking there.

varify/static/scripts/javascript/src/templates.js
```diff
@@ -164,6 +164,7 @@ define([
     var condensedFlags = function(attrs) {
         var flags = [];
         flags.push(['dbSNP', attrs['rsid'] != null]);
+        flags.push(['ClinVar', attrs['solvebio']['clinvar'].length > 0]);
```
naegelyd: I think this should probably be conditional. If `attrs['solvebio']` is not present then it means there is no solvebio integration happening on the server(see my comment on `SOLVEBIO_API_KEY` below). If there is no solvebio callout being made then we will never have clinvar data so we can just omit the flag altogether. @bruth thoughts?

varify/variants/resources.py
```diff
@@ -36,11 +38,17 @@ def get(self, request, pk):
...
+        # also compile the HGVS_c values for the SolveBio query
+        hgvs_c_values = set()
         for eff in data['effects']:
             effects.add(eff['type'])
             if eff.get('transcript') and eff['transcript'].get('gene'):
                 if eff['transcript']['gene']:
                     genes.add(eff['transcript']['gene']['symbol'])
+                if eff['transcript'].get('transcript') and eff.get('hgvs_c'):
+                    hgvs_c_values.add(
+                        '%s:%s'
```
bruth: Using the new-style string formatting is preferred, e.g. `'{0}:{1}'.format(...)`

varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -518,6 +551,7 @@ define([
             $row2.append(this._span(this.renderCohorts(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
             $row2.append(this._span(this.renderFrequencies(attrs), 3));
             $row2.append(this._span(this.renderPubmed(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
+            $row2.append(this._span(this.renderClinvar(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
```
naegelyd: I think appending the Clinvar section should also be conditional on the presence of `attrs.solvebio`(see comment above).

varify/variants/resources.py
```diff
@@ -36,11 +38,17 @@ def get(self, request, pk):
         # this is quite important
         genes = set()
         effects = set()
+        # also compile the HGVS_c values for the SolveBio query
+        hgvs_c_values = set()
         for eff in data['effects']:
```
bruth, line 44: Please add a line between parent statements and `for` and `if` blocks to increase readability. The only exception is the first line within a block such as line 46 just below, but 48 should have a line above it.

varify/variants/resources.py
```diff
@@ -69,6 +77,21 @@ def get(self, request, pk):
 
         data['cohorts'] = cohort_list
 
+        # SolveBio integration
+        # use position, gene symbol, and HGVS notation as possible filters
+        clinvar_filters = RangeFilter(variant.chr, variant.pos, variant.pos)
+        if genes:
```
bruth: Again, one line space as well as for the next `if` block on 86.

varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -386,6 +386,39 @@ define([
             return content.join('');
         },
 
+        _renderClinVarCollection: function(assertions) {
+            var assertion, content, _i, _len;
+            content = [];
+            for (_i = 0, _len = assertions.length; _i < _len; _i++) {
```
naegelyd: Consider removing unnecessary `_len` variable here and replacing `_i` with `i` using `for (i = 0; i < assertions.length; i++)` instead. Mostly a preference thing but that will match the rest of the codebase much more closely.

varify/static/scripts/javascript/src/templates.js
bruth: Yes, I agree. Better to check if `solvebio` is defined first.

varify/static/scripts/javascript/src/ui/modals/result.js
```diff
@@ -519,6 +556,10 @@ define([
             $row2.append(this._span(this.renderFrequencies(attrs), 3));
             $row2.append(this._span(this.renderPubmed(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
 
+            if (attrs['solvebio']) {
+              $row2.append(this._span(this.renderClinVar(attrs), 3).addClass('expandable-details-item').append(this._renderExpandCollapse));
```
naegelyd: The indent here should be 4 spaces, currently it is 2.

PR #207, opened 2014-04-02
Add docs about SolveBio to README and better treat SolveBio as an optional dependency

davecap:
This also fixes a bug that would lead to JS errors when ClinVar queries fail.


naegelyd:
Awesome, thanks Dave! Would you mind squashing the re-run commit into the previous one so that this PR is a single commit? Other than that it looks good and I will merge it after the squash.

PR #220, opened 2014-04-04
TestCases for assessment 

hassanNS:
Tests cases for GET, POST and PUT api calls from resources.py of the assessment app.

### DIFF
--- a/tests/cases/assessments/tests.py
+++ b/tests/cases/assessments/tests.py
@@ -1,0 +1,165 @@
+import json
+import os
+from restlib2.http import codes
+from varify.assessments.resources import *  # noqa
+from ..base import AuthenticatedBaseTestCase
+from django.test.utils import override_settings
+from varify.assessments.models import Assessment, Pathogenicity,\
+    ParentalResult, AssessmentCategory
+from varify.samples.models import Sample, Result
+from django.core import management
+from django_rq import get_worker
+from django.core.cache import cache
+from django_rq import get_queue, get_connection
+from rq.queue import get_failed_queue
+
+TESTS_DIR = os.path.join(os.path.dirname(__file__), '../..')
+SAMPLE_DIRS = [os.path.join(TESTS_DIR, 'samples', 'batch1')]
+
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class AssessmentResourceTestCase(AuthenticatedBaseTestCase):
+
+    def setUp(self):
+        cache.clear()
+        get_queue('variants').empty()
+        get_queue('default').empty()
+        get_failed_queue(get_connection()).empty()
+
+        # Immediately validates and creates a sample
+        management.call_command('samples', 'queue')
+
+        # Synchronously work on queue
+        worker1 = get_worker('variants')
+        worker2 = get_worker('default')
+
+        # Work on variants...
+        worker1.work(burst=True)
+
+        # Work on effects...
+        worker2.work(burst=True)
+
+        # Create and record some data that will be used to create knowledge
+        # capture assessments later on.
+        self.pathogenicity = Pathogenicity(name='pathogenic')
+        self.pathogenicity.save()
+        self.parental_result = ParentalResult(name='heterozygous')
+        self.parental_result.save()
+        self.category = AssessmentCategory(name='other')
+        self.category.save()
+        super(AssessmentResourceTestCase, self).setUp()
+
+    # Test the get method of class AssessmentsResources
+    def test_get_all(self):
+        response = self.client.get('/api/assessments/',
+                                   HTTP_ACCEPT='application/json')
+        self.assertEqual(response.status_code, codes.no_content)
+
+    # Test get method with empty assessment
+    def test_get_when_empty(self):
+        response = self.client.get('/api/assessments/1/',
+                                   HTTP_ACCEPT='application/json')
+        self.assertEqual(response.status_code, codes.not_found)
+
+        sample_id = Sample.objects.get(batch__name='batch1', name='NA12891').id
+        # Create an assessment, then try another get call
+        sample_result = Result.objects.filter(sample_id=sample_id)[0]
+        assessment = Assessment(sample_result=sample_result,
+                                user=self.user,
+                                assessment_category=self.category,
+                                sanger_requested=True,
+                                pathogenicity=self.pathogenicity,
+                                father_result=self.parental_result,
+                                mother_result=self.parental_result)
+        assessment.save()
+
+        response = self.client.get('/api/assessments/1/',
+                                   HTTP_ACCEPT='application/json')
+
+        self.assertEqual(response.status_code, codes.ok)
+        self.assertTrue(response.content)
+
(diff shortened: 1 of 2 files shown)


naegelyd:
Nice, there are a couple changes that will need to be made to the code and to your branch. Commits 1-3 are duplicates of 4-6. Also, 3 and 6 have already been merged into master. You can remove the last commit(the 'merge' one) and instead rebase your branch against master.

```
git fetch upstream
git rebase upstream/master
```

The only files that should be committed here are the test files. All the empty directories, which will eventually have tests, can be removed. I will make line notes on the test code but the commits will need to be cleaned up and made into a single commit. The only relevant commit is the one titled "TestCases for assessments testing get post and put calls" above.

naegelyd:
So, there were a bunch of line notes there but they all follow a common theme. We follow the conventions laid out in [PEP8](http://legacy.python.org/dev/peps/pep-0008/). While we don't require the tests to conform to this code yet because there was too much legacy test code to update, new test code should. To check your code, you can use [flake8](https://flake8.readthedocs.org/en/2.0/). This should show you just the issues in the new test you added:

``` bash
pip install flake8
flake8 tests/cases/assessments/tests.py
```

Most of the output there will be related to the notes I made on the code above. Resolving those issues should keep the code in line with the conventions we are trying to follow in Varify.

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,167 @@
...
+from django_rq import get_worker, get_queue, get_connection
+from rq.queue import get_failed_queue
+
+TESTS_DIR = os.path.join(os.path.dirname(__file__), '../..')
+SAMPLE_DIRS = [os.path.join(TESTS_DIR, 'samples', 'batch1')]
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class AssessmentResourceTestCase(AuthenticatedBaseTestCase):
+
+	def setUp(self):
```
naegelyd: Indent should be 4 spaces, not 8 or a tab character. Note, I am mentioning this here but it applies to the entire file.

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,167 @@
...
+		
+		response = self.client.get('/api/assessments/1/', HTTP_ACCEPT='application/json')
+
+		self.assertEqual(response.status_code, codes.ok)
+		self.assertTrue(response.content)
+
+	# Test post method
+	def test_post_all(self):
+		sample_id = Sample.objects.get(batch__name='batch1', name='NA12891').id
+		# First create an assessment object in json format
```
naegelyd: This comment does not apply to the line below, consider removing altogether

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,167 @@
...
+		sample_result = Result.objects.filter(sample_id=sample_id)[0]
+		count_before = 	Assessment.objects.count()
+		
+		#create a json to pass in
```
naegelyd: All comments should be of the form `# Comment text here...`. Note the space after the `#` and the capitalization of the first letter.

naegelyd:
Add empty `models.py` and `__init__.py` to `tests/cases/assessments/` folder so that the tests can find the `tests.cases.assessments` module.

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,165 @@
...
+        response = self.client.get('/api/assessments/1/',
+                                   HTTP_ACCEPT='application/json')
+
+        self.assertEqual(response.status_code, codes.ok)
+        self.assertTrue(response.content)
+
+    # Test post method
+    def test_post_all(self):
+        sample_id = Sample.objects.get(batch__name='batch1', name='NA12891').id
+        # First create an assessment object in json format
```
naegelyd, line 85: None of the lines adjacent to this comment create an assessment object, consider removing this comment

hassanNS:
will do that now

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,165 @@
...
+        # Get id for the put request and get count
+        a_id = str(assessment.id)
+        count_before = Assessment.objects.count()
+
+        # Create a new category to change our assessment
+        new_category = AssessmentCategory(name='dangerous')
+        new_category.save()
+
+        # Make the json object with the same values (except for the category)
+        assessment_obj = {"sample_result": assessment.id,
```
naegelyd, line 139: You don't need to pass the full object here. If you pass just `"assessment_category": new_category.id` it should keep everything else the same and just update the category.

hassanNS:
ok it's up

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # Now retrive the same object
+        response = self.client.get('/api/assessments/{0}/'.format(a_id),
+                                   HTTP_ACCEPT='application/json')
+
+        # The category should have changed to dangerous
+        if 'dangerous' not in response.content:
+            raise AssertionError("object changed!")
+        # Count should remain the same
+        self.assertEquals(count_before, count_after)
```
naegelyd, line 165: Move this up in the code closer to the assignment of `count_after` so it is easier to follow what is going on.

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,165 @@
...
+        # Tests, object should be changed, count should be same
+        self.assertEqual(response.status_code, codes.ok)
+
+        # Now retrive the same object
+        response = self.client.get('/api/assessments/{0}/'.format(a_id),
+                                   HTTP_ACCEPT='application/json')
+
+        # The category should have changed to dangerous
+        if 'dangerous' not in response.content:
+            raise AssertionError("object changed!")
```
naegelyd, line 163: I'm not sure I follow the logic behind this. Shouldn't this just be:

``` python
# The category should have changed to dangerous
assessment = json.loads(response.content)
self.assertEqual(assessment.assessment_category.name, 'dangerous')
```

I don't think there is a need to raise an exception here, just fail the test if the value is not what we expect.

tests/cases/assessments/tests.py
```diff
@@ -0,0 +1,165 @@
...
+        # Post should be successful and new object should be added
+        self.assertEqual(response.status_code, codes.created)
+        self.assertEqual(count_after, count_before+1)
+
+        # Testing post with bad value
+        response = self.client.post('/api/assessments/',
+                                    data=json.dumps({}),
+                                    content_type='application/json',
+                                    HTTP_ACCEPT='application/json')
+        # Should be unprocessable
```
naegelyd, line 113: A lot of these comments are trivial in the sense that they just describe exactly what the line does. For example, the next line should be easily understood by anyone looking at the code. The comment doesn't explain anything particularly confusing or add any significant clarity or instruction to the reader. Consider removing this comment and any others that are explaining very obvious code or very general such as the `# Testing PUT call` comment below.

tests/cases/assessments/tests.py
hassanNS, line 163: wow! ok that's a much better way

PR #225, opened 2014-04-04
Expose SolveBio exceptions to the integration in VariantResource

davecap:
Gives better visibility into SolveBio query errors without crashing requests for variant details (requires upgrade to the latest `solvebio` Python client 1.2.0).

### DIFF
--- a/varify/variants/resources.py
+++ b/varify/variants/resources.py
@@ -15,12 +15,11 @@ from .models import Variant
 log = logging.getLogger(__name__)
 
 try:
-    from solvebio.contrib.django_solvebio.client import client \
-        as solvebio_client
-    from solvebio import Filter, RangeFilter  # noqa
+    from solvebio.contrib.django_solvebio import SolveBio
+    from solvebio import SolveError, Filter, RangeFilter  # noqa
 except ImportError:
-    solvebio_client = None
-    log.warning('Could not import the SolveBio client')
+    SolveBio = None
+    log.warning('Could not import SolveBio')
 
 
 class VariantResource(ThrottledResource):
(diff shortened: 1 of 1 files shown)


varify/variants/resources.py
```diff
@@ -100,14 +99,16 @@ def get(self, request, pk):
...
-            # only add the 'clinvar' key if the query succeeds
-            if clinvar is not None:
-                data['solvebio']['clinvar'] = clinvar
+            try:
+                # get the ClinVar Dataset resource by its alias
+                clinvar = SolveBio.get_dataset('clinvar')
+                # list() ensures the query is executed within the try/except
+                results = list(clinvar.query(filters=filters))
+            except Exception as e:
+                log.error('SolveBio ClinVar query failed: {0}'.format(str(e)))
```
bruth: Prefer a `log.exception` as it will automatically capture the exception details. This is useful for log handlers (such as Sentry) that takes the exception details and displays them in the UI.

bruth: The message can be left as is.

varify/variants/resources.py
```diff
@@ -100,14 +99,16 @@ def get(self, request, pk):
...
             else:
-                log.error('SolveBio ClinVar query failed')
+                # only add the 'clinvar' key if the query succeeds
+                data['solvebio']['clinvar'] = results
```
bruth: This can be moved up to the `try` block as:

``` python
data['solvebio']['clinvar'] = list(clinvar.query(filters=filters))
```

No need for the `else`. As a side note, does the SolveBio API use a custom base exception class to differentiate it's thrown exceptions?

davecap: Yea we use a `SolveError` exception class, it may make more sense to catch that instead.

bruth:
Looks good. For future reference, please read [our style guide](https://github.com/cbmi/style-guides/blob/master/git.md#commit-messages) for Git commit messages. The first line should be direct and terse. Additional information should be put in the body of the message.

Any final thoughts @naegelyd?

naegelyd:
Build is failing due to flake8 issue. That should be fixed in master and then we should rebase this against master

naegelyd:
I'm just going to merge this. The broken build is not a result of changes on this branch. Thanks for these changes, I'll test them Monday

PR #235, opened 2014-04-08
Fix cohort popover sample links

### LINKED ISSUE #205: Clicking sample name in popover modal does not change sample

naegelyd:
Fix #205.


varify/static/js/src/ui/sample.js
```diff
@@ -0,0 +1,100 @@
...
+            else {
+                this.showError('There was an issue parsing the sample ID. ' +
+                               'Please make sure it is valid and try again');
+            }
+        },
+
+        showError: function(text) {
+            this.ui.loadingMessage.hide();
+            this.ui.errorMessage.text(text);
+            this.ui.errorMessage.show();
```
bruth: The `.show()` can be chained on the above statement.

varify/static/js/src/ui/sample.js
```diff
@@ -0,0 +1,100 @@
...
+
+        onSampleFetchSuccess: function() {
+            // Since this page relies on the user clicking a link on the
+            // results page, there must have already been a sample selected. If
+            // not, show an error message.
+            var json = _.clone(this.data.context.get('json')),
+                foundSample = false;
+
+            if (json && json.children && json.children.length) {
+                for (var i = 0; i < json.children.length; i++) {
```
bruth, line 71: Using `_.any` would be cleaner here since it breaks once found.

naegelyd, line 71: I understand the argument for `_.any` but I think this looping method is easier to follow. If you feel really strongly about `_.any` I will move it over to that but I am in favor of keeping this as is because I think it reads clearest this way.

bruth, line 71: Not required, just a suggestion.

PR #241, opened 2014-04-09
Optimize the SolveBio ClinVar query to prevent slow variant resource requests

davecap:
Remove the by-gene filter to narrow down the result sets and return only the first 10 results from the query.
Adds the total result count to the `clinvar` dictionary in `VariantResource`. Also fixes an issue with the genomic position filter.


varify/static/js/src/templates.js
```diff
@@ -168,8 +168,8 @@ define([
         flags.push(['1000g', attrs['1000g'].length > 0]);
         flags.push(['EVS', attrs['evs'].length > 0]);
 
-        if (!! attrs['solvebio'] && !! attrs['solvebio']['clinvar']) {
-            flags.push(['ClinVar', attrs['solvebio']['clinvar'].length > 0]);
+        if (!! attrs.solvebio && !! attrs.solvebio.clinvar) {
```
bruth: I don't see a need to coerce these expressions to `true` values.

varify/variants/resources.py
```diff
@@ -89,22 +89,29 @@ def get(self, request, pk):
...
             data['solvebio'] = {}
 
-            # ClinVar integration -- use position, gene symbol, and HGVS
-            # notation as possible filters.
-            filters = RangeFilter(variant.chr, variant.pos, variant.pos)
-
-            if genes:
-                filters = filters | Filter(gene_symbol__in=list(genes))
+            # ClinVar integration -- use position and HGVS
+            filters = Filter(chromosome=str(variant.chr),
```
bruth: Although `str(variant.chr)` works, it is safer to do `variant.chr.value`.

varify/variants/resources.py
```diff
@@ -89,22 +89,29 @@ def get(self, request, pk):
...
+                q = SolveBio.get_dataset('clinvar').query(
+                    limit=10,  # limit to 10 results (single page)
+                    filters=filters)
+                # Send the first page of results to the client
+                data['solvebio']['clinvar'] = {
+                    'results': q.results,
+                    'total': q.total
+                }
             except SolveError as e:
                 log.exception('SolveBio ClinVar query failed: {}'.format(e))
```
bruth: Not part of the PR, but just noticed this. Python 2.6 requires positional string placeholders e.g. `{}` to have the index defined. So this should be changed to `{0}`.

PR #243, opened 2014-04-10
Only truncate once before updating frequencies

### LINKED ISSUE #238: Cohort frequencies for non-world cohort are not being recalculated
When the `allele-freqs` command is run, the only cohort that has its frequencies updated is the `World` cohort. All the `CohortVariant` records are linked to the `World` cohort.

naegelyd:
Fix #238.

This fixes a bug where we were only seeing the frequencies being updated for the last cohort in the list. This was a result of the placement of the truncate command. Previously, the truncation happened every loop iteration. Now, it happens before the loop starts.


bruth:
Oh my.. that is quite a bug.

PR #246, opened 2014-04-10
Create test for allele frequency calculation

ryanjohara:

### DIFF
--- a/tests/cases/commands/tests.py
+++ b/tests/cases/commands/tests.py
@@ -660,3 +660,28 @@ class DeleteTestCase(QueueTestCase):
         self.assertEqual(0, Sample.objects.count())
         self.assertEqual(0, Batch.objects.count())
         self.assertEqual(0, Project.objects.count())
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class AlleleTestCase(QueueTestCase):
+    def setUp(self):
+        super(AlleleTestCase, self).setUp()
+
+        # Immediately validates and creates a sample
+        management.call_command('samples', 'queue')
+
+        # Synchronously work on queue
+        worker1 = get_worker('variants')
+        worker2 = get_worker('default')
+
+        # Work on variants...
+        worker1.work(burst=True)
+
+        # Work on effects...
+        worker2.work(burst=True)
+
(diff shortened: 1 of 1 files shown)


tests/cases/commands/tests.py
```diff
@@ -662,3 +662,28 @@ def test_delete(self):
...
+        worker1.work(burst=True)
+
+        # Work on effects...
+        worker2.work(burst=True)
+
+    def test_allele_freqs(self):
+        management.call_command('samples', 'allele-freqs')
+        firstCount = CohortVariant.objects.count()
+        management.call_command('samples', 'allele-freqs')
+        secondCount = CohortVariant.objects.count()
```
naegelyd: No need for this assignment, just use:

``` python
self.assertEqual(firstCount, CohortVariant.objects.count())
```

tests/cases/commands/tests.py
```diff
@@ -662,3 +662,28 @@ def test_delete(self):
...
+
+        # Work on variants...
+        worker1.work(burst=True)
+
+        # Work on effects...
+        worker2.work(burst=True)
+
+    def test_allele_freqs(self):
+        management.call_command('samples', 'allele-freqs')
+        firstCount = CohortVariant.objects.count()
```
naegelyd, line 684: This should check that the cohort variant count is non-zero. If `firstCount` and `secondCount` are both 0 then we have no idea that the truncation is working since there was nothing to truncate.

tests/cases/commands/tests.py
```diff
@@ -660,3 +660,30 @@ def test_delete(self):
...
+        self.assertNotEqual(firstCount, 0)
+        management.call_command('samples', 'allele-freqs')
+        secondCount = CohortVariant.objects.count()
+        self.assertNotEqual(secondCount, 0)
```
naegelyd: If we have reached this point, we know that `firstCount != 0`. Therefore, the below `self.assertEqual(firstCount, secondCount)` can NEVER be true if `secondCount == 0`. Remove the above `self.assertNotEqual(secondCount, 0)` and just use `self.assertEqual(firstCount, CohortVariant.objects.count())` below.

ryanjohara: That is correct.  Change has been made.

PR #262, opened 2014-04-18
Update delete-samples command to take multiple ids

### LINKED ISSUE #261: Update delete-sample command to take multiple sample ids

naegelyd:
Fix #261.

The ids are simply arguments after the command, each separated by a space. For example, if we wanted to delete samples with IDs 4 though 8 we could do:

`./bin/manage.py samples delete-sample 4 5 6 7 8`

### DIFF
--- a/varify/samples/management/subcommands/delete_sample.py
+++ b/varify/samples/management/subcommands/delete_sample.py
@@ -22,12 +22,11 @@ class Command(BaseCommand):
         database = options.get('database')
         cursor = connections[database].cursor()
 
-        assessments = Assessment.objects.filter(
-            sample_result__sample_id__in=args)
+        invalid_ids = set(Assessment.objects.filter(
+            sample_result__sample_id__in=args).values_list(
+            'sample_result__sample_id', flat=True))
 
-        invalid_ids = []
-        if assessments.exists():
-            invalid_ids = assessments.values_list('id', flat=True)
+        if invalid_ids:
             log.warning("Sample(s) with id '{0}' cannot be deleted "
                         "because they have knowledge capture data "
                         "associated with them. If you are absolutely "
--- a/varify/samples/management/subcommands/delete_sample.py
+++ b/varify/samples/management/subcommands/delete_sample.py
@@ -19,55 +19,65 @@ class Command(BaseCommand):
         if not args:
             raise CommandError('A sample ID must be specified')
 
-        sample_id = args[0]
-
-        if Assessment.objects.filter(
-                sample_result__sample_id=sample_id).exists():
-            print("This sample cannot be deleted because it has knowledge "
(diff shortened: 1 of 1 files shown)


bruth:
Is there a reason you are not using an `IN ( ... )` clause rather than looping over each sample?

naegelyd:
I need to loop over each sample to check it for assessments so I figured I'd just leave the code the same and delete per sample too. I suppose I could first cull all the IDs that have assessment data then do a mass delete on those. I will update the PR.

varify/samples/management/subcommands/delete_sample.py
```diff
@@ -19,55 +19,65 @@ def handle(self, *args, **options):
...
+            log.warning("Sample(s) with id '{0}' cannot be deleted "
+                        "because they have knowledge capture data "
+                        "associated with them. If you are absolutely "
+                        "sure you want to delete these samples, "
+                        "manually delete the knowledge capture data "
+                        "for them and then rerun this command."
+                        .format(", ".join([str(id) for id in invalid_ids])))
+
+        valid_ids = \
+            ", ".join([str(id) for id in args if id not in invalid_ids])
```
bruth: Looks like a classic SQL injection mistake... the string passed into `format()` should be the bind parameters and the ids themselves can be passed as arguments.

varify/samples/management/subcommands/delete_sample.py
```diff
@@ -19,55 +19,65 @@ def handle(self, *args, **options):
...
-                  "command.")
-            return
-
         database = options.get('database')
         cursor = connections[database].cursor()
 
+        assessments = Assessment.objects.filter(
+            sample_result__sample_id__in=args)
+
+        invalid_ids = []
```
bruth: Rather than doing an `exists()` followed by a `values_list()` (two queries), it would be simpler to set `invalid_ids` to the distinct list of ids and check if it's not empty. `invalid_ids = set(assessments.values_list('id', flat=True).distinct())`.

varify/samples/management/subcommands/delete_sample.py
```diff
@@ -19,55 +19,64 @@ def handle(self, *args, **options):
...
-                  "capture data associated with it. If you are absolutely "
-                  "sure you want to delete this sample, manually delete the "
-                  "knowledge capture data for this sample and then rerun this "
-                  "command.")
-            return
-
         database = options.get('database')
         cursor = connections[database].cursor()
 
+        invalid_ids = set(Assessment.objects.filter(
```
bruth: How about, `set(Sample.objects.filter(results__assessment__isnull=False).values_list('id', flat=True))`

bruth: (and `pk__in=args`)

PR #265, opened 2014-04-21
Fix for automatically starting workers if not started

### LINKED ISSUE #260: Queue command should check for workers
If the variants and default rqworkers are not found to be running when the queue command runs then it should log a warning and then start the workers itself.

hassanNS:
Fix #260
Checks if rqworkers are already started. If not, outputs a warning and starts rq's automatically.

### DIFF
--- a/tests/cases/sample_load_process/tests.py
+++ b/tests/cases/sample_load_process/tests.py
@@ -167,13 +167,8 @@ class SnpeffReloadTest(QueueTestCase):
     def test(self):
         "Load a single VCF, reload the snpEff data using the same VCF."
         management.call_command('samples', 'queue',
-                                os.path.join(SAMPLE_DIRS[0], 'batch1/locus_1'))
-
-        # Synchronously work on queue
-        worker1 = get_worker('variants')
-        worker2 = get_worker('default')
-        worker1.work(burst=True)
-        worker2.work(burst=True)
+                                os.path.join(SAMPLE_DIRS[0], 'batch1/locus_1'),
+                                burst=True, startworkers=True)
 
         expected_variant_effects_count = 5426
 
--- a/varify/samples/management/subcommands/queue.py
+++ b/varify/samples/management/subcommands/queue.py
@@ -1,6 +1,10 @@
 import os
+import redis
 import glob
 import logging
+from django_rq import get_worker
+from rq import Worker
+from redis.exceptions import ConnectionError
 from optparse import make_option
 from django.conf import settings
 from django.core.management.base import BaseCommand
@@ -20,6 +24,16 @@ class Command(BaseCommand):
         make_option('--max', action='store', dest='max', default=None,
                     type='int',
                     help='Specifies the maximum number of samples to queue.'),
+        make_option('--burst', action='store_true', dest='burst',
+                    default=False,
+                    help='Include --burst flag to run workers in burst mode. '
+                         'Used in conjuction with --startworkers. This flag '
+                         'has no effect if --startworkers is not present.'),
+        make_option('--startworkers', action='store_true', dest='startworkers',
+                    default=False,
+                    help='Include --startworkers flag to automatically start '
+                         'the variant and default workers if they are not '
+                         'already running. '),
     )
 
     def _queue(self, dirs, max_count, database, verbosity):
@@ -73,10 +87,56 @@ class Command(BaseCommand):
 
         return count, scanned
 
+    def startWorkers(self, burst):
+        # Find the number of current workers
+        queues = getattr(settings, 'RQ_QUEUES', {})
+        default = queues['default'] if 'default' in queues else None
+        variants = queues['variants'] if 'variants' in queues else None
+
+        if not (queues and default and variants):
+            log.warning('RQ_QUEUES settings could not be found')
+            return
+
+        # Create connections to redis to identify the workers
+        def_connection = redis.Redis(host=default['HOST'],
+                                     port=default['PORT'],
+                                     db=default['DB'])
+        var_connection = redis.Redis(host=variants['HOST'],
+                                     port=variants['PORT'],
+                                     db=variants['DB'])
+
+        # Get all the workers connected with our redis server
+        try:
+            all_workers = Worker.all(def_connection) + \
+                Worker.all(var_connection)
+        except ConnectionError:
+            log.warning('Could not connect to redis server to create workers. '
+                        'Please make sure Redis server is running')
+            return
+
+        found_default = False
+        found_variant = False
+
+        # Loop through all the workers (even duplicates)
+        for worker in all_workers:
+            found_default = found_default or 'default' in worker.queue_names()
+            found_variant = found_variant or 'variants' in worker.queue_names()
+
+        # Start the required worker
+        if not found_variant:
+            log.debug('Did not find variants worker. Starting ... ')
+            get_worker('variants').work(burst=burst)
+
+        if not found_default:
+            log.debug('Did not find default worker. Starting ... ')
+            get_worker('default').work(burst=burst)
+
     def handle(self, *dirs, **options):
         database = options.get('database')
         max_count = options.get('max')
         verbosity = int(options.get('verbosity'))
+        burst = options.get('burst')
+        workers = options.get('startworkers')
 
         if not dirs:
             dirs = []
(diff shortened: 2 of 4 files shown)


naegelyd:
The README should be updated as part of this pull request if it needs changing.

varify/samples/management/subcommands/queue.py
```diff
@@ -1,15 +1,17 @@
...
 from django.conf import settings
+from django.core.management import call_command
 from django.core.management.base import BaseCommand
 from django.db import DEFAULT_DB_ALIAS
 from varify.samples.pipeline.handlers import load_samples
 
 SAMPLE_DIRS = getattr(settings, 'VARIFY_SAMPLE_DIRS', ())
 
-log = logging.getLogger(__name__)
+log = logging.getLogger(__file__)
```
naegelyd: This was changed to `__name__` in master recently and should not be changed back to `__file__`.

varify/samples/management/subcommands/queue.py
```diff
@@ -73,6 +75,36 @@ def _queue(self, dirs, max_count, database, verbosity):
 
         return count, scanned
 
+    def startWorkers(self):
+        # Parse output of running rqinfo to check which workers are running
+        output = subprocess.check_output(['rqinfo', '-W', '-i', '0'])
```
naegelyd: Instead of using `rqinfo`, I think it would be better to read the workers themselves using something like this example:

``` python
>>> import rq
>>> import redis
>>> from rq import Worker
>>> from django.conf import settings
>>> queues = getattr(settings, 'RQ_QUEUES', {})
>>> default = queues['default']
>>> variants = queues['variants']
>>> default_connection = connection = redis.Redis(host=default['HOST'], port=default['PORT'], db=default['DB'])
KeyboardInterrupt
>>> default_connection = redis.Redis(host=default['HOST'], port=default['PORT'], db=default['DB'])
>>> variants_connection = redis.Redis(host=variants['HOST'], port=variants['PORT'], db=variants['DB'])
>>> default_workers = Worker.all(default_connection)
...
>>> variants_workers = Worker.all(variants_connection)
...
>>> all_workers = default_workers + variants_workers
```

Where the `...` is all the workers using that connection. Then, you can do something like 

``` python
found_default = False
found_variants = False
for w in all_workers:
    found_default = found_default or 'default' in w.queue_names()
    found_variants = found_variants or 'variants' in w.queue_names()

if not found_default:
    // start default worker
if not found_variants:
   // start varaint worker
```

hassanNS:
It seems call_command will not work in the case of starting two workers. The first command hangs up and doesn't go to the next call_command

varify/samples/management/subcommands/queue.py
naegelyd: Note, in our case, the HOST, DB, and PORT are the same for both but since there is no guarantee that that will be the case in all cases I tried to account for any differences above by treating default and variants queues separately.

hassanNS:
the changes should be up now

naegelyd:
The queue command should be updated to take a `burst` argument that will be passed to the worker when it is created. This way, if the user is loading a single sample they can burst the workers by passing burst to the queue command.

naegelyd:
Once the above comment is incorporated, then the tests should be updated to remove all the worker manipulation and just use the queue command with the burst argument.

hassanNS:
oops! Build failing due to flake8 issue. Will fix now

hassanNS:
In order to burst the worker, I'm guessing the command would be 
./bin/manage.py rqworker --burst [worker] & ?

naegelyd:
@muslimgiga Yes, passing the `--burst` flag like you have above will do the trick.

tests/cases/sample_load_process/tests.py
```diff
@@ -167,13 +167,8 @@ class SnpeffReloadTest(QueueTestCase):
...
         management.call_command('samples', 'queue',
-                                os.path.join(SAMPLE_DIRS[0], 'batch1/locus_1'))
-
-        # Synchronously work on queue
-        worker1 = get_worker('variants')
-        worker2 = get_worker('default')
-        worker1.work(burst=True)
-        worker2.work(burst=True)
+                                os.path.join(SAMPLE_DIRS[0], 'batch1/locus_1'),
+                                burst=True, startworker=True,)
```
hassanNS: incorrect argument passed in

naegelyd:
The commit history seems to have gotten out of whack. The only commit should be the one with comment: `Automatically start rqworkers, updated tests`

hassanNS:
I tried rebasing against upstream master and for some reason it squashed your delete_sample commit with mine

hassanNS:
@naegelyd I see what you mean by not using `bursting`, the argument could be passed right in. 
Will update these

varify/samples/management/subcommands/queue.py
```diff
@@ -73,10 +82,52 @@ def _queue(self, dirs, max_count, database, verbosity):
...
+        # Create connections to redis to identify the workers
+        def_connection = redis.Redis(host=default['HOST'],
+                                     port=default['PORT'],
+                                     db=default['DB'])
+        var_connection = redis.Redis(host=variants['HOST'],
+                                     port=variants['PORT'],
+                                     db=variants['DB'])
+
+        # Get all the workers connected with our redis server
+        all_workers = Worker.all(def_connection) + Worker.all(var_connection)
```
hassanNS: @naegelyd surrounding this line with a try catch to catch a ConnectionError seems to be the way to check for the redis server

hassanNS:
Code updated

varify/samples/management/subcommands/queue.py
```diff
@@ -20,6 +24,18 @@ class Command(BaseCommand):
         make_option('--max', action='store', dest='max', default=None,
                     type='int',
                     help='Specifies the maximum number of samples to queue.'),
+        make_option('--burst', action='store_true', dest='burst',
+                    default=False,
+                    help='Include --burst flag to run workers in burst \
```
naegelyd: The `\` and `\n` can all be removed since this is between parenthesis. I believe this should work fine:

``` python
make_option('--startworkers', action='store_true', dest='startworkers',
            default=False,
            help='Include --startworkers flag to automatically start the '
                 'variant and default workers if they are not already running.')
```

varify/samples/management/subcommands/queue.py
```diff
@@ -73,10 +89,56 @@ def _queue(self, dirs, max_count, database, verbosity):
...
+        var_connection = redis.Redis(host=variants['HOST'],
+                                     port=variants['PORT'],
+                                     db=variants['DB'])
+
+        # Get all the workers connected with our redis server
+        try:
+            all_workers = Worker.all(def_connection) + \
+                Worker.all(var_connection)
+        except ConnectionError:
+            log.warning('Could not connect to redis server to create workers.\
```
naegelyd: See note above about string concatenation. The reason it works in between parens is that string literals next to each other are automatically concatenated(see [here](https://docs.python.org/2/tutorial/introduction.html#strings)).

varify/samples/management/subcommands/queue.py
```diff
@@ -20,6 +24,18 @@ class Command(BaseCommand):
...
+                    help='Include --burst flag to run workers in burst \
+                            \nmode. Used in conjuction with --startworkers. \
+                            \nThis flag has no effect if --startworkers is \
+                            \nnot present.'),
+        make_option('--startworkers', action='store_true', dest='startworkers',
+                    default=False,
+                    help='Include --startworkers flag to automatically \
+                            \nstart the variant and default workers \
+                            \nif they are not already \
+                            \nrunning.'),
```
naegelyd: See note above about string literals

hassanNS:
wow this makes things so much easier thanks!

varify/samples/management/subcommands/queue.py
```diff
@@ -20,6 +24,18 @@ class Command(BaseCommand):
         make_option('--max', action='store', dest='max', default=None,
                     type='int',
                     help='Specifies the maximum number of samples to queue.'),
+        make_option('--burst', action='store_true', dest='burst',
+                    default=False,
+                    help='Include --burst flag to run workers in burst'
```
naegelyd: Most of these lines are missing a space at the end. Note, when the concatenation happens, nothing is used as a separator so we usually like to end lines with a space in order to get the space between words and so that all the lines begin with a character.

varify/samples/management/subcommands/queue.py
```diff
@@ -20,6 +24,18 @@ class Command(BaseCommand):
...
+                         'mode. Used in conjuction with --startworkers.'
+                         'This flag has no effect if --startworkers is'
+                         'not present.'),
+        make_option('--startworkers', action='store_true', dest='startworkers',
+                    default=False,
+                    help='Include --startworkers flag to automatically'
```
naegelyd: These lines get progressively shorter, move as many words as possible to line above to avoid exceeding per-line character limits.

hassanNS:
Updated

PR #266, opened 2014-04-21
Hide cohort popover tab changes or modal is closed

### LINKED ISSUE #264: Cohort popover remains forever visible when modal closed before cohort hidden
Steps to recreate:
1. Open the variant details modal by clicking on any of the result rows
2. Click on a cohort to show the popover
3. Click the close button
4. Click another result
5. At this point, a cohort popover(the one resulting from `2`) will be visible but should have been closed and destroyed when the modal itself was closed. 

Clicking new cohorts after `5` will result in 2 cohort popovers being visible.

naegelyd:
Fix #264.


varify/static/js/src/ui/modals/result.js
```diff
@@ -30,7 +33,12 @@ define([
         },
 
         hidePopover: function(event) {
-            $('.cohort-sample-popover').not(event.target).popover('hide');
+            if (event && event.target) {
+                $('.cohort-sample-popover').not(event.target).popover('hide');
```
bruth, line 39: I realize this is out of the scope of this PR, but using a _global_ selector here is a bit fragile. I am not exactly what the better approach is, either the popover listening to a standard event or the common parent delegating to the popover, but I am just making a note. No changes need to be made at this time.

varify/static/js/src/ui/modals/result.js
```diff
@@ -30,7 +33,12 @@ define([
         },
 
         hidePopover: function(event) {
```
bruth, line 37: Although `this` is not being accessed here, this method should be added to the `_.bindAll` call.

varify/static/js/src/ui/modals/result.js
naegelyd, line 39: Yea, this is just one of many issues with this code. I am planning a large scale refactor most of the JS but thanks for pointing this out.

PR #275, opened 2014-04-28
Changed tests to use fixtures, fixed bug w/ gene resource

hassanNS:
Updated PR moving all resource test cases into folder `resources`. The PR on master can be closed as the commit history is behind.


hassanNS:
it seems the problem running the assessment tests persists

PR #277, opened 2014-04-29
Implement SampleDialog for new method of selecting a sample

### LINKED ISSUE #2: Refactor proband selection screen
The proband selection screen is messy with the 3 drop down menus. The reality is that the batch is useful for internal tracking, but our users don't know what it means. It will become even less relevant as more people start loading data. I would suggest a control similar to the Harvest ID filter view used for concepts with lots of values. There should be a search bar at the top that allows one to search by project or sample ID. Below that, there should be a paginated table that shows: Sample ID, Project (batch), and Date Loaded. Users should be able to select a single row from this table and apply it to their query
image

bruth:
Fix #2


varify/migrations/0004_remove_sample_as_queryable.py
```diff
@@ -0,0 +1,216 @@
...
+        "Write your forwards methods here."
+        orms['DataConcept'].objects\
+                .filter(app_name='samples', model_name='sample', field_name='id')\
+                .update(queryable=False)
+
+    def backwards(self, orm):
+        "Write your backwards methods here."
+        orms['DataConcept'].objects\
+                .filter(app_name='samples', model_name='sample', field_name='id')\
+                .update(queryable=True)
```
naegelyd, line 19: I hope that we don't need to run this backward migration but this currently would cause some issues. Under the current setup, there are 2 sample concepts. One is viewable(not queryable) and contains sample, project, and batch and one is queryable(not viewable) and contains just contains sample. The reason for this is that there isn't(wasn't?) a way to have fields in a concept but exclude them from queries and keep controls for those fields from appearing on the query page. I really do not think this is a major issue but if we do migrate backwards here, we will need to go in and manually change one of the sample concepts to just be viewable.

An alternate approach to this migration would be to remove the queryable sample concept in the forward migration and create it in the backward migration.

bruth:
Screenshot of the dialog:

screen_shot

varify/samples/resources.py
```diff
@@ -37,33 +38,44 @@ def sample_posthook(instance, data, request):
...
 
     @api.cache_resource
     def get(self, request, pk):
-        try:
-            sample = self.model.objects.select_related(
-                'batch', 'project').get(pk=pk)
-        except self.model.DoesNotExist:
-            raise Http404
-
+        instance = self.get_object(request, pk=pk)
```
naegelyd, line 72: Do we need to call get_object again here or can we just read the instance as `request.instance`?

varify/migrations/0004_remove_sample_as_queryable.py
bruth, line 19: If the record is removed, then all the client code that depends on the concept ID would break. I agree this is a tad grimy and shows some unfortunate coupling, but I am not sure of an alternate means.

varify/samples/resources.py
bruth, line 72: I did that more for transparency. The fact that it's being stored on `request` is an implementation detail (to prevent a second lookup), but `get_object` _should_ be used. This is how most of Serrano does it as well.

varify/samples/resources.py
```diff
@@ -251,19 +278,19 @@ class PhenotypeResource(ThrottledResource):
...
-        if recalculate == "true":
+        if recalculate == 'true':
             try:
                 management.call_command('samples', 'gene-ranks', sample_id,
                                         force=True)
             except Exception:
-                log.exception("Error recalculating gene rankings")
-                return HttpResponse("Error recalculating gene rankings",
+                log.exception('Error recalculating gene rankings')
+                return HttpResponse('Error recalculating gene rankings',
```
naegelyd: Consider using [codes](https://github.com/bruth/restlib2/blob/master/restlib2/http/__init__.py#L16) and replacing HttpResponse with render as was done [here](https://github.com/cbmi/serrano/pull/169). This might be outside the scope of this PR though so separate issues will need to be created if that is the case.

varify/static/js/src/ui/modals/sample.js
```diff
@@ -0,0 +1,312 @@
...
+
+        applySort: function(attr) {
+            var dir = 'asc';
+
+            // Already sorted by the attribute, cycle direction
+            if (this.collection._sortAttr === attr) {
+                dir = this.collection._sortDir === 'asc' ? 'desc' : 'asc';
+            }
+
+            var p;
```
naegelyd: Move down closer to first use? Also, consider giving this a more descriptive name or a comment describing what it is.

varify/static/stylesheets/scss/style.scss
```diff
@@ -283,3 +283,66 @@ body, #concept-panel, #context-panel {
...
+            font-size: 14px;
+        }
+
+        button {
+            float: right;
+        }
+    }
+
+    table.table {
+
```
naegelyd: Remove blank line

PR #278, opened 2014-04-30
Migration bug fix

### LINKED ISSUE #268: DeserializationError during initial migration
When running `syncdb --migrate` the following error is seen:

```
> varify:0002_avocado_metadata_migration
> varify:0003_fix_sequence_problems
- Loading initial data for varify.
Problem installing fixture 'varify/fixtures/initial_data.json': Traceback (most recent call last):
  File "/Users/ohara/varify-env/lib/python2.7/site-packages/django/core/management/commands/http://loaddata.py", line 190, in handle
    for obj in objects:
  File "/Users/ohara/varify-env/lib/python2.7/site-packages/django/core/serializers/http://json.py", line 47, in Deserializer
    raise DeserializationError(e)
DeserializationError: ContentType matching query does not exist.
```

hassanNS:
Fix #268
Problem as described by @naegelyd 
`The issue was that permissions and content types for models in an app are automatically created int he DB when the app is created. The initial_data has natural keys for content types in the permissions list referring to the samples app so that is why it needs to be first`
Signed-off-by: Sheik Hassan solergiga@yahoo.com

### DIFF
--- a/varify/conf/global_settings.py
+++ b/varify/conf/global_settings.py
@@ -24,6 +24,7 @@ INSTALLED_APPS = (
     'django.contrib.sites',
     'django.contrib.staticfiles',
 
+    'varify.samples',
     'varify',
 
     'varify.raw',
@@ -34,7 +35,6 @@ INSTALLED_APPS = (
     'varify.genes',
     'varify.variants',
     'varify.phenotypes',
-    'varify.samples',
     'varify.literature',
 
     'varify.support',
@@ -58,6 +58,7 @@ INSTALLED_APPS = (
     'modeltree',
     'guardian',
     'reversion',
+
 )
 
 


naegelyd:
A little more detail...The [initial_data](https://github.com/cbmi/varify/blob/da9d4bc84974e94859427c88d18fe9cff85643ab/varify/fixtures/initial_data.json) file contains a group called 'Sample Managers' with a list of permissions(in natural key form) related to models in the samples app. The natural key for a Permission has [2 parts](https://github.com/django/django/blob/e1812617cf534b00230e38af118194d8bf0b48b6/django/contrib/auth/models.py#L84-L85): the codename and the [natural key of the ContentType](https://github.com/django/django/blob/e1812617cf534b00230e38af118194d8bf0b48b6/django/contrib/contenttypes/models.py#L168-L169). The ContentType key is made up of the app and model name. The DeserializationError occurred during the intial data load because it tried to reference a ContentType for an app and model that did not exist. Moving `varify.samples` above `varify` in the list of installed apps forces the ContentTypes and Permissions to be created in the DB before they are referenced in the initial data of the varify application.

naegelyd:
Had to back this change out as varify.samples relies on avocado and the initial migrations were failing.

PR #285, opened 2014-05-01
Refactor result details modal

naegelyd:
This removes the knowledge capture components from the modal and also removes all the HTML construction from the JS in favor of using Marionette views and HTML templates.


varify/static/js/src/ui/modals/result.js
```diff
@@ -3,531 +3,13 @@
...
-
-                content.push('</ul>');
-            }
-
-            return content.join('');
-        },
+    '../variant'
+], function($, _, Backbone, Marionette, utils, variant) {
 
+    /*
```
bruth: Should this commented out block still be here?

varify/static/js/src/ui/modals/result.js
```diff
@@ -768,84 +79,44 @@ define([
...
-            'click #variant-details-link': 'hideButtons',
-            'click #knowledge-capture-link': 'knowledgeCaptureTabClicked',
-            'click [data-target=expand-collapse-link]': 'toggleExpandedState'
-        },
-
-        initialize: function() {
-            _.bindAll(this, 'onSaveError', 'onSaveSuccess');
-
-            this.assessmentTab = new AssessmentTab;
+        selectors: {
```
bruth: Is this a new convention you are starting for selectors external to the view? A comment should be added to make note of this.

bruth: Ideally, in the future the coupling should be refactoring even through events or some parent view.

naegelyd: I only did this out of convenience and for clarify. I cannot use the standard `ui` convention because the actual elements are not yet available even though they exist in the `el`. That is, if I tried to put all the `selectors` into the `ui` block like is normal convention, then lines like this in `open` fail because the ui element is empty:

``` javascript
...
ui: {
    rows: '[data-target=expandable-row]'
},
...
open: fucntion() {
    ...
    this.ui.rows.css('height', '' + this.maxExpandableHeight + 'px')
        .css('overflow', 'hidden');
    ...
```

I'm not trying to start a new trend or anything. I was just trying to group all the selectors I was using into a single property for clarity. If there is a better approach, just let me know and I will update to use it instead. Otherwise, I will just add a comment to explain the appraoch to the code itself.

varify/static/js/src/ui/modals/result.js
```diff
@@ -896,67 +116,94 @@ define([
...
+            }));
+
+            this.cohorts.show(new variant.Cohorts({
+                collection: new Backbone.Collection(
+                    this.model.get('variant').cohorts
+                )
+            }));
+
+            this.frequencies.show(new variant.Frequencies({
+                model: new Backbone.Model(this.model.get('variant'))
```
bruth: Is there any value in initializing a single model for the `variant`? This is being used a couple times above. The `Variant` model could define all the sub-{models, collections} for the nested structures such cohorts, etc. This makes it a bit more clear what embedded data is supported and enables implementing sub-document parsing on nested models/collections. For example:

``` javascript
var Variant = Backbone.Model.extend({
    constructor: function(attrs, options) {
        // Ensure to parse..
        options = _.defaults({parse: true}, options);

        this.cohorts = new Backbone.Collection();
        this.frequencies = new Backbone.Collection();
        // etc..

        Backbone.Model.prototype.constructor.call(this, attrs, options);
    },
    parse: function(attrs) {
        this.frequencies.reset(attrs.frequencies);
        this.cohorts.reset(attrs.cohorts);
        // etc..
        return attrs;
    }
});
```

bruth: Each of this sub-{models, collections} would correspond to the views they are being passed into.

varify/static/js/src/ui/modals/result.js
```diff
@@ -896,67 +116,94 @@ define([
...
-
-            this.assessmentTab.update(assessmentModel);
+            this.assessmentMetrics.show(metricsView);
 
             this.$el.modal('show');
 
             // Reset the row and item heights and overflow styles as they may
             // have been toggled previously.
-            $('[data-target=expandable-details-row]').css('height', '' + this.maxExpandableHeight + 'px')
+            this.$el.find(this.ui.expandableRows)
```
bruth: This looks to be redundant since `this.ui.expandableRows` is already a jQuery object and is scoped to `this.$el`.

varify/static/js/src/ui/modals/result.js
```diff
@@ -896,67 +116,94 @@ define([
...
+                .css('height', '' + this.maxExpandableHeight + 'px')
                 .css('overflow', 'hidden');
-            $('[data-target=expand-collapse-link]').text(this.showMoreText);
+            this.$el.find(this.selectors.expandCollapseLink)
```
bruth: Should this be _found_ in this element? Isn't the point of the `this.selectors` hash to be external to this element?

varify/static/js/src/ui/tables/row/genes.js
```diff
@@ -0,0 +1,29 @@
...
+    var EmptyGene = Marionette.ItemView.extend({
+        template: 'varify/tables/row/genes/empty'
+    });
+
+    var GeneItem = Marionette.ItemView.extend({
+        template: 'varify/tables/row/genes/item'
+    });
+
+    var GeneList = Marionette.CollectionView.extend({
+        template: function() {},
```
bruth: `CollectionView`s do not need a `template` defined.

varify/static/js/src/models/variant.js
```diff
@@ -0,0 +1,52 @@
...
+
+            this.phenotypes.reset(utils.groupPhenotypesByType(attrs));
+            this.cohorts.reset(attrs.cohorts);
+            this.articles.reset(utils.groupArticlesByType(attrs));
+
+            var clinvarResults;
+            if (attrs.solvebio && attrs.solvebio.clinvar) {
+                clinvarResults = attrs.solvebio.clinvar.results;
+            }
+            clinvarResults = clinvarResults || [];
```
bruth: Small nit, but `reset` does not require an array to be passed.

varify/static/js/src/ui/modals/result.js
```diff
@@ -896,67 +113,85 @@ define([
...
 
             this.$el.modal('show');
 
+            // Since the UI elements are bound before rendering, many of the
+            // UI elements we need access to to will not be found at this point
+            // as they did not exist in the templates and therefore would not
+            // be found during the normal UI binding call. We rebind here so
+            // since all the elements should now be available as all the
+            // regions have been populated and their views shown.
+            this.bindUIElements();
```
bruth, line 187: Do any of the selectors depends on collection view items? If so, could there be a race condition with the items rendering before these elements are rebound?

varify/static/templates/variant/clinvar-item.html
```diff
@@ -0,0 +1,12 @@
+Assertion: <a target='_blank' href='https://www.ncbi.nlm.nih.gov/clinvar/'<%= data.assertion.rcvaccession %>/'><%= data.assertion.rcvaccession %></a>
+
+<ul>
+    <li>
+        <small>Siginificance</small> <b><%= data.assertion.clinicalsignificance %></b>
```
bruth: Significance.

varify/static/templates/variant/clinvar-item.html
```diff
@@ -0,0 +1,12 @@
+Assertion: <a target='_blank' href='https://www.ncbi.nlm.nih.gov/clinvar/'<%= data.assertion.rcvaccession %>/'><%= data.assertion.rcvaccession %></a>
```
bruth: I am getting a complaint in the console that `data.assertion` is undefined.

varify/static/js/src/ui/modals/result.js
naegelyd, line 187: No. None of the selectors depend on collection view items. They are all in the templates of the different regions themselves so there should be no race condition here.

varify/static/js/src/ui/variant/assessment-metrics.js
```diff
@@ -0,0 +1,256 @@
...
+        itemViewContainer: '[data-target=items]',
+
+        ui: {
+            popover: '[data-target=user-popover]',
+        },
+
+        initialize: function() {
+            _.bindAll(this, 'hidePopover');
+
+            $(document).on('click', this.hidePopover);
```
bruth: This should go in `onRender` since the handler depends on UI elements to be rendered.

varify/static/js/src/ui/variant/assessment-metrics.js
```diff
@@ -0,0 +1,256 @@
...
+        itemViewContainer: '[data-target=items]',
+
+        ui: {
+            popover: '[data-target=details-popover]',
+        },
+
+        initialize: function() {
+            _.bindAll(this, 'hidePopover');
+
+            $(document).on('click', this.hidePopover);
```
bruth: Same as above, put in `onRender`.

varify/static/js/src/ui/variant/assessment-metrics.js
```diff
@@ -0,0 +1,256 @@
...
+        onClose: function() {
+            // Clean up after ourselves. If we don't remove the event handler
+            // then we will still receive click notifications even long after
+            // we are dead and gone and the hidePopover method will throw a
+            // fit because ui.popover is no longer bound.
+            $(document).off('click', this.hidePopover);
+        },
+
+        onRender: function() {
+            this.bindUIElements();
```
bruth, line 86: What is the purpose of doing this explicitly?

varify/static/js/src/ui/variant/assessment-metrics.js
```diff
@@ -0,0 +1,256 @@
...
```
bruth, line 154: Same as above, why do this explicitly?

varify/static/js/src/ui/variant/assessment-metrics.js
```diff
@@ -0,0 +1,256 @@
...
+        },
+
+        initialize: function() {
+            _.bindAll(this, 'onError', 'onRequest', 'onSync');
+
+            if (!this.options.variantId) {
+                throw new Error('Variant ID Required');
+            }
+
+            this.model= new models.AssessmentMetrics({}, {
```
bruth: Space before `=`. `null` can be passed to the model rather than initializing an object.

varify/static/js/src/ui/variant/cohorts.js
```diff
@@ -0,0 +1,107 @@
...
+        itemViewContainer: '[data-target=items]',
+
+        ui: {
+            popover: '[data-target=cohort-popover]',
+        },
+
+        initialize: function() {
+            _.bindAll(this, 'hidePopover');
+
+            $(document).on('click', this.hidePopover);
```
bruth: Put in `onRender`.

varify/static/js/src/ui/variant/cohorts.js
```diff
@@ -0,0 +1,107 @@
...
+        onClose: function() {
+            // Clean up after ourselves. If we don't remove the event handler
+            // then we will still receive click notifications even long after
+            // we are dead and gone and the hidePopover method will throw a
+            // fit because ui.popover is no longer bound.
+            $(document).off('click', this.hidePopover);
+        },
+
+        onRender: function() {
+            this.bindUIElements();
```
bruth, line 94: Remove assuming it's redundant from the above questions.

varify/static/js/src/ui/variant/assessment-metrics.js
naegelyd, line 86: The popovers are not part of this view's template but are instead on the itemviews. Because of that, the initial UI binding will not find the popovers. So, we need to bind the UI elements after the itemviews have rendered so that the popovers will be found.

varify/static/js/src/ui/variant/effects.js
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            var gene, transcript = this.model.get('transcript'),
+                uniqueGenes = this.model.get('uniqueGenes');
+
+            if (uniqueGenes && uniqueGenes.length && transcript) {
+                gene = transcript.gene;
+            }
+
+            var proteinChange, changeString =
```
bruth: Reading this was a tad deceiving, consider moving `changeString` to next line with it's assignment.

varify/static/js/src/ui/variant/phenotypes.js
```diff
@@ -0,0 +1,89 @@
...
+        message: 'No phenotypes for this gene'
+    });
+
+    var PhenotypeItem = Marionette.ItemView.extend({
+        template: 'varify/variant/phenotype-item',
+
+        tagName: 'li',
+
+        serializeData: function() {
+            var data = Marionette.ItemView.prototype.serializeData.apply(
```
bruth, line 32: All this does it call `toJSON()` on the model or collection that is bound. You can leave it if you want.

PR #287, opened 2014-05-02
Resolved issue with deleting queries

### LINKED ISSUE #268: DeserializationError during initial migration

hassanNS:
Fix #268
The issue correcting `main.js` to register the deleteQuery function.

Signed-off-by: Sheik Hassan solergiga@yahoo.com

### DIFF
--- a/varify/static/js/src/main.js
+++ b/varify/static/js/src/main.js
@@ -179,6 +179,8 @@ require({
                     collection: this.data.queries
                 }),
 
+                deleteQuery: new c.ui.DeleteQueryDialog(),
+
                 resultDetails: new ui.ResultDetails,
 
                 phenotype: new ui.Phenotype({


varify/static/js/src/main.js
```diff
@@ -179,6 +179,12 @@ require({
                     collection: this.data.queries
                 }),
 
+                deleteQuery: new c.ui.DeleteQueryDialog({
```
naegelyd: DeleteQueryDialog does not require any options. See [Cilantro's registration](https://github.com/cbmi/cilantro/blob/master/src/js/cilantro/main.js#L70) as the example.

PR #289, opened 2014-05-05
Convert cohort_variant and result primary key types to bigint

### LINKED ISSUE #222: Convert potentially enormous tables to use bigint data types
sample_result and possibly the cohort frequency tables will run out of primary keys at around 2 billion records because django defaults to int4 for its primary keys. We need to research what approach other people have used and test this REALLY well prior to deploying. A quick stack overflow search shows that there are pitfalls to some approaches, so this is a measure twice, cut once situation.

ryanjohara:
Backwards migration is NOT supported.
MySQL, Oracle, and SQLite are NOT supported.
ONLY forward migration in PostgreSQL is supported.
Resolves #222

### DIFF
--- a/tests/cases/commands/tests.py
+++ b/tests/cases/commands/tests.py
@@ -11,6 +11,7 @@ from varify.assessments.models import Assessment, Pathogenicity, \
     ParentalResult, AssessmentCategory
 from varify.samples.models import Sample, CohortSample, Result, SampleRun, \
     SampleManifest, Project, Cohort, Batch, CohortVariant, ResultScore
+from varify.variants.models import Variant
 from varify.samples.management.subcommands import gene_ranks
 from ..sample_load_process.tests import QueueTestCase
 from ...models import MockHandler
@@ -666,3 +667,41 @@ class AlleleTestCase(QueueTestCase):
         self.assertNotEqual(firstCount, 0)
         management.call_command('samples', 'allele-freqs')
         self.assertEqual(firstCount, CohortVariant.objects.count())
+
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class BigAutoFieldsTestCase(TestCase):
+    def setUp(self):
+        self.project = Project(id=1, name='', label='')
+        self.project.save()
+        self.batch = Batch(id=1, name='', project_id=self.project.id, label='')
+        self.batch.save()
+
+    def test_big_auto_fields(self):
+        # Test that BigAutoFields can exceed Django's default integer limit
+        result_count_before = Result.objects.count()
+        cohort_variant_count_before = CohortVariant.objects.count()
+        two_hundred_billion = int(200e9)
+
+        new_sample = Sample.objects.create(
+            id=1, label='',batch_id=self.batch.id, version=1,
+            count=1, published=True, name='', project_id=self.project.id)
+        
+        new_variant = Variant.objects.create(
+            id=1, chr_id=1, pos=1, ref='', alt='', md5='')
+
+        new_result = Result.objects.create(
+            id=two_hundred_billion, sample_id=new_sample.id,
+            variant_id=new_variant.id)
+
+        new_score = ResultScore.objects.create(
+            id=1, result_id=new_result.id, rank=0, score=0)
+        
+        new_cohort_variant = CohortVariant.objects.create(
+            id=two_hundred_billion, variant_id=new_variant.id, cohort_id=1)
+        
+        result_count_after = Result.objects.count()        
+        cohort_variant_count_after = CohortVariant.objects.count()
+    
+        self.assertEqual(result_count_before+1, result_count_after)
+        self.assertEqual(cohort_variant_count_before+1, cohort_variant_count_after)
--- a/varify/core/models.py
+++ b/varify/core/models.py
@@ -1,4 +1,7 @@
+from django.core import exceptions
 from django.db import models
+from django.db.models import fields
+from south.modelsinspector import add_introspection_rules
 
 
 class TimestampedModel(models.Model):
@@ -28,3 +31,61 @@ class LabeledModel(models.Model):
 
     def __unicode__(self):
         return self.label
+
+
+class BigIntegerField(fields.IntegerField):
+    def db_type(self, connection):
+        if 'postgres' in connection.__class__.__module__:
+            return 'bigint'
+        else:
+            raise NotImplemented
+
+    def get_internal_type(self):
+        return 'BigIntegerField'
+
+    def to_python(self, value):
+        if value is None:
+            return value
+        try:
+            return long(value)
+        except (TypeError, ValueError):
+            raise exceptions.ValidationError(
+                'This value must be a long integer.')
+
+
+class BigAutoField(fields.AutoField):
+    # ONLY PostgreSQL is supported (MySQL, Oracle & SQLite are NOT supported)
+    def db_type(self, connection):
+        if 'postgres' in connection.__class__.__module__:
+            return 'bigserial'
+        else:
+            raise NotImplemented
+
+    def get_internal_type(self):
+        return 'BigAutoField'
+
+    def to_python(self, value):
+        if value is None:
+            return value
+        try:
+            return long(value)
+        except (TypeError, ValueError):
+            raise exceptions.ValidationError(
+                'This value must be a long integer.')
+
+
+class BigForeignKey(fields.related.ForeignKey):
+    def db_type(self, connection):
+        rel_field = self.rel.get_related_field()
+        is_bigauto = isinstance(rel_field, BigAutoField)
+        is_bigint = isinstance(rel_field, BigIntegerField)
+        matches_type = connection.features.related_fields_match_type
+        if (is_bigauto or (not matches_type and is_bigint)):
+
+            return BigIntegerField().db_type(connection)
+        return rel_field.db_type(connection)
+
+
+add_introspection_rules([], ["^varify\.core\.models\.BigIntegerField"])
+add_introspection_rules([], ["^varify\.core\.models\.BigAutoField"])
+add_introspection_rules([], ["^varify\.core\.models\.BigForeignKey"])
--- a/varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
+++ b/varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
@@ -1,0 +1,285 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+class Migration(SchemaMigration):
+    # Backwards migration is NOT supported.
+    # MySQL, Oracle, and SQLite are NOT supported.
+    # ONLY forward migration in PostgreSQL is supported.
+    def forwards(self, orm):
+        # Changing field 'CohortVariant.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "cohort_variant" ALTER COLUMN "id" TYPE bigint')
+
+        # Changing field 'Result.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "sample_result" ALTER COLUMN "id" TYPE bigint')
+        # Change FK fields 'result_score.result_id' and 'assessment.sample_result_id'
+        db.execute('ALTER TABLE "result_score" ALTER COLUMN "result_id" TYPE bigint')
+        db.execute('ALTER TABLE "assessment" ALTER COLUMN "sample_result_id" TYPE bigint')
+
+    def backwards(self, orm):
+        raise RuntimeError("Cannot convert bigint to integer")
+
+    models = {
+        'auth.group': {
+            'Meta': {'object_name': 'Group'},
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'unique': 'True', 'max_length': '80'}),
+            'permissions': ('django.db.models.fields.related.ManyToManyField', [], {'to': "orm['auth.Permission']", 'symmetrical': 'False', 'blank': 'True'})
+        },
+        'auth.permission': {
+            'Meta': {'ordering': "('content_type__app_label', 'content_type__model', 'codename')", 'unique_together': "(('content_type', 'codename'),)", 'object_name': 'Permission'},
+            'codename': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'content_type': ('django.db.models.fields.related.ForeignKey', [], {'to': "orm['contenttypes.ContentType']"}),
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'max_length': '50'})
+        },
+        'auth.user': {
+            'Meta': {'object_name': 'User'},
+            'date_joined': ('django.db.models.fields.DateTimeField', [], {'default': 'datetime.datetime.now'}),
+            'email': ('django.db.models.fields.EmailField', [], {'max_length': '75', 'blank': 'True'}),
+            'first_name': ('django.db.models.fields.CharField', [], {'max_length': '30', 'blank': 'True'}),
+            'groups': ('django.db.models.fields.related.ManyToManyField', [], {'to': "orm['auth.Group']", 'symmetrical': 'False', 'blank': 'True'}),
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'is_active': ('django.db.models.fields.BooleanField', [], {'default': 'True'}),
+            'is_staff': ('django.db.models.fields.BooleanField', [], {'default': 'False'}),
+            'is_superuser': ('django.db.models.fields.BooleanField', [], {'default': 'False'}),
+            'last_login': ('django.db.models.fields.DateTimeField', [], {'default': 'datetime.datetime.now'}),
+            'last_name': ('django.db.models.fields.CharField', [], {'max_length': '30', 'blank': 'True'}),
+            'password': ('django.db.models.fields.CharField', [], {'max_length': '128'}),
+            'user_permissions': ('django.db.models.fields.related.ManyToManyField', [], {'to': "orm['auth.Permission']", 'symmetrical': 'False', 'blank': 'True'}),
+            'username': ('django.db.models.fields.CharField', [], {'unique': 'True', 'max_length': '30'})
+        },
+        'contenttypes.contenttype': {
+            'Meta': {'ordering': "('name',)", 'unique_together': "(('app_label', 'model'),)", 'object_name': 'ContentType', 'db_table': "'django_content_type'"},
+            'app_label': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'model': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'name': ('django.db.models.fields.CharField', [], {'max_length': '100'})
+        },
+        'genome.chromosome': {
+            'Meta': {'ordering': "['order']", 'object_name': 'Chromosome', 'db_table': "'chromosome'"},
+            'code': ('django.db.models.fields.IntegerField', [], {'null': 'True'}),
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'label': ('django.db.models.fields.CharField', [], {'max_length': '2'}),
+            'order': ('django.db.models.fields.IntegerField', [], {'default': '0'}),
+            'value': ('django.db.models.fields.CharField', [], {'max_length': '2', 'db_index': 'True'})
+        },
+        'genome.genome': {
+            'Meta': {'object_name': 'Genome', 'db_table': "'genome'"},
+            'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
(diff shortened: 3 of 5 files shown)


tests/cases/commands/tests.py
```diff
@@ -666,3 +667,45 @@ def test_allele_freqs(self):
...
+
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class BigAutoFieldsTestCase(QueueTestCase):
+    def setUp(self):
+        super(BigAutoFieldsTestCase, self).setUp()
+
+        # Immediately validates and creates a sample
+        management.call_command('samples', 'queue',
+                                burst=True, startworkers=True)
```
naegelyd: Is there any reason for calling this command? You seem to be manually creating the sample and other data below.

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,45 @@ def test_allele_freqs(self):
         self.assertNotEqual(firstCount, 0)
         management.call_command('samples', 'allele-freqs')
         self.assertEqual(firstCount, CohortVariant.objects.count())
+
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class BigAutoFieldsTestCase(QueueTestCase):
+    def setUp(self):
+        super(BigAutoFieldsTestCase, self).setUp()
```
naegelyd: If there is no reason for the command below, then remove this whole `setUp` method

tests/cases/commands/tests.py
ryanjohara: After removing that line of code, the following error occurs when running the BigAutoField test:

<code>IntegrityError: insert or update on table "sample" violates foreign key constraint "sample_project_id_fkey"
DETAIL:  Key (project_id)=(1) is not present in table "project".</code>

naegelyd: I would remove the command and do something like:

``` python
@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
class BigAutoFieldsTestCase(TestCase):
    def setUp(self):
        self.project = Project.create(...)
        self.batch = Batch.create(project_id=self.project.id, ...)

# Project and batch were created in setup so their IDs can be
# used when creating other objects below by referencing:
#    self.project.id       or      self.batch.id

....
```

naegelyd: Note the change from `QueueTestCase` to `TestCase`

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,45 @@ def test_allele_freqs(self):
...
+        management.call_command('samples', 'queue',
+                                burst=True, startworkers=True)
+
+    def test_big_auto_fields(self):
+        # Test that BigAutoFields can exceed Django's default integer limit
+        resultCountBefore = Result.objects.count()
+        cohortVariantCountBefore = CohortVariant.objects.count()
+
+        newSample = Sample(
+            id=2000000000, label='',batch_id=1,version=1,
```
naegelyd: You are testing with an id of `2,000,000,000` but the maximum for a normal integer is `2,147,483,647` as listed [here](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.IntegerField) so it is not actually exceeding the limit that these changes are intended to increase. Increase this to 200 trillion or something else obnoxiously large to prove the point. Also, since it seems this same number is used as the ID throughout the tests, consider storing this in a variable above then referencing that variable throughout these tests.

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Changing field 'Result.id'
+        #db.alter_column('sample_result', 'id', self.gf('varify.core.models.BigAutoField')(primary_key=True))
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "sample_result" ALTER COLUMN "id" TYPE bigint')
+        # Change FK fields 'result_score.result_id' and 'assessment.sample_result_id'
+        db.execute('ALTER TABLE "result_score" ALTER COLUMN "result_id" TYPE bigint')
+        db.execute('ALTER TABLE "assessment" ALTER COLUMN "sample_result_id" TYPE bigint')
+    def backwards(self, orm):
+
```
naegelyd: The comments above make it very clear that backwards migration is not supported but this attempts to do one anyway. Remove the code here and instead refuse to backwards migrate and warn the user. Try something like the below(also, see [this migration](https://github.com/cbmi/varify/blob/0069e45d7542b50daa1b82548d039655cf892622/varify/assessments/migrations/0002_auto__chg_field_assessment_category.py#L18) for another example:

``` python

def backwards(self, orm):
    raise RuntimeError("EXPLAIN WHY MIGRATION CAN'T BE REVERSED")
```

varify/core/models.py
```diff
@@ -28,3 +30,21 @@ class Meta(object):
 
     def __unicode__(self):
         return self.label
+
+
+class BigAutoField(fields.AutoField):
+    # ONLY PostgreSQL is supported (MySQL, Oracle & SQLite are NOT supported)
```
naegelyd, line 57: Since we are only supporting postgres, shouldn't `NotImplemented` be raised in all cases where `'postgres' in connection.__class__.__module__` is False?

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,299 @@
...
+
+class Migration(SchemaMigration):
+    # Backwards migration is NOT supported.
+    # MySQL, Oracle, and SQLite are NOT supported.
+    # ONLY forward migration in PostgreSQL is supported.
+    def forwards(self, orm):
+
+        # Changing field 'CohortVariant.id'
+        #db.alter_column('cohort_variant', 'id', self.gf('varify.core.models.BigAutoField')(primary_key=True))
```
naegelyd: Remove commented line

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,299 @@
...
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "cohort_variant" ALTER COLUMN "id" TYPE bigint')
+
+        # Changing field 'Result.id'
+        #db.alter_column('sample_result', 'id', self.gf('varify.core.models.BigAutoField')(primary_key=True))
```
naegelyd: Remove commented line

naegelyd:
I see from the migration that sample_result, cohort_varaint, result_score, and assessment are all updated to use bigint. However, the only models that I see being changed are Result and CohortVariant. As noted above, I don't think we need to modify assessment at all. The ResultScore model will need to be updated. Currently, the table is altered in the migration but the model is unchanged.

ryanjohara:
Thanks for reviewing. I believe all of the comments have been addressed.

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
         self.assertNotEqual(firstCount, 0)
         management.call_command('samples', 'allele-freqs')
         self.assertEqual(firstCount, CohortVariant.objects.count())
+
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class BigAutoFieldsTestCase(TestCase):
+    def setUp(self):
+        self.project = Project(id=1,name='',label='')
+        self.project.save()
```
naegelyd, line 676: This is a nit but you can do:

``` python
self.project = Project.objects.create(id=1,name='',label='')
```

instead of:

``` python
self.project = Project(id=1,name='',label='')
self.project.save()
```

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+        self.batch.save()
+
+    def test_big_auto_fields(self):
+        # Test that BigAutoFields can exceed Django's default integer limit
+        resultCountBefore = Result.objects.count()
+        cohortVariantCountBefore = CohortVariant.objects.count()
+        twoHundredBillion = 200000000000
+
+        newSample = Sample(
+            id=1, label='',batch_id=self.batch.id,version=1,
```
naegelyd: Add spaces after commas

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+            count=1, published=True, name='',project_id=self.project.id
+            )
+        newSample.save()
+        
+        newVariant = Variant(
+            id=1,chr_id=1,pos=1,ref='',alt='',md5=''
```
naegelyd: Spaces after commas

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+    def test_big_auto_fields(self):
+        # Test that BigAutoFields can exceed Django's default integer limit
+        resultCountBefore = Result.objects.count()
+        cohortVariantCountBefore = CohortVariant.objects.count()
+        twoHundredBillion = 200000000000
+
+        newSample = Sample(
+            id=1, label='',batch_id=self.batch.id,version=1,
+            count=1, published=True, name='',project_id=self.project.id
+            )
```
naegelyd: Move this to previous line

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+        newSample.save()
+        
+        newVariant = Variant(
+            id=1,chr_id=1,pos=1,ref='',alt='',md5=''
+            )
```
naegelyd: move paren to previous line

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+        newVariant.save()
+
+        newResult = Result(
+            id=twoHundredBillion,sample_id=newSample.id,
```
naegelyd: space after commas

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,51 @@ def test_allele_freqs(self):
...
+            variant_id=newVariant.id
+            )
```
naegelyd: For everything below, move the single parens to the previous line and add spaces after the commas

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+        return "BigIntegerField"
+
+    def to_python(self, value):
+        if value is None:
+            return value
+        try:
+            return long(value)
+        except (TypeError, ValueError):
+            raise exceptions.ValidationError(
+                "This value must be a long integer."
```
naegelyd: Use single quotes

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+                )
```
naegelyd: Move paren to previous line

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+
+class BigIntegerField(fields.IntegerField):
+    def db_type(self, connection):
+        if 'postgres' in connection.__class__.__module__:
+            return 'bigint'
+        else:
+            raise NotImplemented
+
+    def get_internal_type(self):
+        return "BigIntegerField"
```
naegelyd: Single quotes

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+class BigAutoField(fields.AutoField):
+    # ONLY PostgreSQL is supported (MySQL, Oracle & SQLite are NOT supported)
+    def db_type(self, connection):
+        if 'postgres' in connection.__class__.__module__:
+            return 'bigserial'
+        else:
+            raise NotImplemented
+
+    def get_internal_type(self):
+        return "BigAutoField"
```
naegelyd: single quotes

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+
+    def to_python(self, value):
+        if value is None:
+            return value
+        try:
+            return long(value)
+        except (TypeError, ValueError):
+            raise exceptions.ValidationError(
+                "This value must be a long integer."
```
naegelyd: single quotes

varify/core/models.py
```diff
@@ -28,3 +31,61 @@ class Meta(object):
...
+                )
```
naegelyd: move to previous line

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,284 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+    # Backwards migration is NOT supported.
+    # MySQL, Oracle, and SQLite are NOT supported.
+    # ONLY forward migration in PostgreSQL is supported.
+    def forwards(self, orm):
+
```
naegelyd: Remove blank line

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,284 @@
...
+        # Changing field 'CohortVariant.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "cohort_variant" ALTER COLUMN "id" TYPE bigint')
+
+        # Changing field 'Result.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "sample_result" ALTER COLUMN "id" TYPE bigint')
+        # Change FK fields 'result_score.result_id' and 'assessment.sample_result_id'
+        db.execute('ALTER TABLE "result_score" ALTER COLUMN "result_id" TYPE bigint')
```
naegelyd, line 20: Add blank line below

varify/samples/migrations/0034_auto__chg_field_cohortvariant_id__chg_field_result_id.py
```diff
@@ -0,0 +1,284 @@
...
+    # ONLY forward migration in PostgreSQL is supported.
+    def forwards(self, orm):
+
+        # Changing field 'CohortVariant.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "cohort_variant" ALTER COLUMN "id" TYPE bigint')
+
+        # Changing field 'Result.id'
+        # Override auto generated migration code
+        db.execute('ALTER TABLE "sample_result" ALTER COLUMN "id" TYPE bigint')
```
naegelyd, line 18: The [assessment model](https://github.com/cbmi/varify/blob/6e57f46b6baf157111e6a8b4bb95ea35c3466e2c/varify/assessments/models.py#L55) also needs to be updated as it has a foreign key reference into Result as well. I have not checked the code but any other models with Result foreign keys will need to be updated as well.

ryanjohara, line 18: OK. I think <code>result_score.result_id</code> and <code>assessment.sample_result_id</code> are the only FKs to <code>sample_result</code>.

varify/core/models.py
```diff
@@ -28,3 +31,59 @@ class Meta(object):
...
+            raise exceptions.ValidationError(
+                'This value must be a long integer.')
+
+
+class BigForeignKey(fields.related.ForeignKey):
+    def db_type(self, connection):
+        rel_field = self.rel.get_related_field()
+        if (isinstance(rel_field, BigAutoField) or
+            (not connection.features.related_fields_match_type and
+                isinstance(rel_field, BigIntegerField))):
```
naegelyd: I am not sure how this can best be cleaned up but this if statement seems very unwieldy as it stands now, would something like the following be better.

``` python
is_bigauto = isinstance(rel_field, BigAutoField)
is_bigint = isinstance(rel_field, BigIntegerField)
matches_type = connection.features.related_fields_match_type

if (is_bigauto or (not matches_type and is_bigint)):
    ...
```

varify/core/models.py
```diff
@@ -28,3 +31,60 @@ class Meta(object):
...
+                'This value must be a long integer.')
+
+
+class BigForeignKey(fields.related.ForeignKey):
+    def db_type(self, connection):
+        rel_field = self.rel.get_related_field()
+        is_bigauto = isinstance(rel_field, BigAutoField)
+        is_bigint = isinstance(rel_field, BigIntegerField)
+        matches_type = connection.features.related_fields_match_type
+        if (is_bigauto or (not matches_type and is_bigint)):
```
naegelyd, line 83: Add blank line above

varify/core/models.py
```diff
@@ -28,3 +31,60 @@ class Meta(object):
...
+            return BigIntegerField().db_type(connection)
+        return rel_field.db_type(connection)
```
naegelyd, line 86: Add blank line above

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,41 @@ def test_allele_freqs(self):
...
+        self.project = Project(id=1, name='', label='')
+        self.project.save()
+        self.batch = Batch(id=1, name='', project_id=self.project.id, label='')
+        self.batch.save()
+
+    def test_big_auto_fields(self):
+        # Test that BigAutoFields can exceed Django's default integer limit
+        resultCountBefore = Result.objects.count()
+        cohortVariantCountBefore = CohortVariant.objects.count()
+        twoHundredBillion = 200000000000
```
bruth: This is a nit, but do you mind changing this to `int(200e9)`.. the number gets a tad blurry after a 6 zeroes.

tests/cases/commands/tests.py
```diff
@@ -666,3 +667,41 @@ def test_allele_freqs(self):
...
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class BigAutoFieldsTestCase(TestCase):
+    def setUp(self):
+        self.project = Project(id=1, name='', label='')
+        self.project.save()
+        self.batch = Batch(id=1, name='', project_id=self.project.id, label='')
+        self.batch.save()
+
+    def test_big_auto_fields(self):
```
bruth, line 680: Our style guide adheres to having Python variables use underscores, not camelcase. Please change the variables below.

naegelyd, line 680: Here's our general style repo: https://github.com/cbmi/style-guides and the Python style guide: http://legacy.python.org/dev/peps/pep-0008/

bruth:
I found this gist: https://gist.github.com/thomasyip/3158388 that seems to support Oracle and SQLite.

naegelyd:
I figured we would just support Postgres since [Avocado currently has issues with Oracle and MySQL](https://github.com/cbmi/avocado/issues/202). But, if that patch works, we might as well try to support as many DB types as possible even if we are only ever using this on Postgres.

ryanjohara:
Code has been added to support MySQL, SQLite, and Oracle.  After a successful syncdb and migrate dry run with SQLite, I encountered the following error when going forward with a real migration:

<code>varify:0003_fix_sequence_problems</code>
<code>FATAL ERROR - The following SQL query failed: alter sequence avocado_dataconcept_id_seq restart with 35;</code>
<code>The error was: near "sequence": syntax error</code>
<code>File "/Users/ohara/varify-env/lib/python2.7/site-packages/django/db/backends/sqlite3/base.py", line 344, in execute</code>
<code>    return Database.Cursor.execute(self, query, params)</code>
<code>django.db.utils.DatabaseError: near "sequence": syntax error</code>

naegelyd:
Hmm, [the migration](https://github.com/cbmi/varify/blob/96fb7e14c1b5993d975da02addf82fa8df128bcd/varify/migrations/0003_fix_sequence_problems.py) will need to be updated to check the DB it is running on. You can set the sqllite sequnce using something like. So, the migration could check if the DB is sqllite and then use the sql below(using replacing the seq# and table_name of course) instead of what is currently in the migration.

``` sql
UPDATE SQLITE_SEQUENCE SET seq = <SEQ_#> WHERE name = '<TABLE_NAME>'
```

mitalia:
Just a comment, as far as Varify is concerned, while it would be nice to have support for other database backends, we already have a fair bit of postgres-specific stuff (for example, using the postgres copy commands) that might need re-tooling. Given that there are significant database design and tuning considerations that need to be made when building something like Varify at scale, I think we need to be realistic that the gap between _technically_ supporting other database engines and _actually_ supporting them is quite wide. If there's an easy thing we can do to be platform-agnostic, that's great. But we shouldn't expend lots of effort on other platforms without a pressing need.

ryanjohara:
That sequence does not appear to exist after running syncdb when using SQLite:

<code>FATAL ERROR - The following SQL query failed: UPDATE avocado_dataconcept_id_seq set seq = 35 WHERE name = 'avocado_dataconcept';</code>
<code>The error was: no such table: avocado_dataconcept_id_seq</code>

Should it be created?

naegelyd:
I believe that the query would be:

```
UPDATE SQLITE_SEQUENCE set seq = 35 WHERE name = 'avocado_dataconcept';
```

naegelyd:
Given the trouble even sqlite is presenting with just the migrations and Mike's point above about `pgcopy` and other postrgres-specific usage in Varify, I say we move forward with this PR only supporting postgres for now and we can address other DB's as needed later on.

@bruth @mitalia Any objections with Ryan reverting this PR a couple steps back to the point where it only supported postgres?

mitalia:
We need it for Postgres. Everything else would be a "nice to have" no issue from me.

bruth:
Revert

ryanjohara:
Only forward migration in PostgreSQL is supported now.

PR #296, opened 2014-05-09
Implement workspace page for sample detail and variant set display

bruth:


bruth:
Screen shot of what it looks like right now.

screen shot 2014-05-09 at 8 55 56 am

varify/samples/resources.py
```diff
@@ -352,24 +348,98 @@ def get(self, request, year, month, day, name):
...
+        sample_result_resource,
+        name='variant'),
+
+    url(r'^(?P<pk>\d+)/variants/sets/$',
+        sample_result_sets_resource,
+        name='variant-sets'),
+
+    url(r'^variants/sets/(?P<pk>\d+)/$',
+        sample_result_set_resource,
+        name='variant-set'),
```
naegelyd, line 431: This is a nit but variant set and result set seem to be used interchangeably here and in the changes above so far. I'd like to agree on a single name and try to use either variant set or result set for consistency's sake. Thoughts?

varify/static/js/src/ui/workflows/workspace.js
```diff
@@ -0,0 +1,77 @@
...
+
+        onRender: function() {
+            c.ui.WorkspaceWorkflow.prototype.onRender.call(this);
+
+            this.ui.sampleVariantSetHelp.tooltip({
+                title: 'A variant set is a fixed set of sample variants that ' +
+                       'can be annotated and augmented at the creator\'s discretion.'
+            });
+
+            // Handlers unbound on close
```
naegelyd: Does this comment mean that they are automatically unbound on close? Just want to check since I don't see the handlers being explicitly unbound anywhere.

varify/samples/resources.py
bruth, line 431: I would prefer to call them 'result sets' in the code (including URLs).. even if we display/call them variant sets in the UI. Sound good?

naegelyd, line 431: Yea, that sounds fine.

varify/static/js/src/ui/workflows/workspace.js
bruth: Yes, since `listenTo` is used.. not sure why I put a comment when it's implied. I will remove it.

varify/samples/resources.py
bruth, line 431: Hmm, the `SampleResultsResource` already uses `/samples/4/variants/`, likewise, `/samples/variants/32323/` for a single _sample variant_. I think the variant sets should be consistent with those (as they are now).

bruth, line 431: Since the model is called `ResultSet`, the resource is named after that. I think of these as _variant sets_ relative to a sample.

naegelyd, line 431: Agreed, after looking this over some more, I think that everything here is fine as is. We already use the split naming convention so these changes agree with what we already use.

PR #297, opened 2014-05-12
Progress for Variant importer from spreadsheet

hassanNS:
The importer resource so far has had a 100% matching with two large spreadsheets.
The javascript and ui portions of the importer are not included

### DIFF
--- a/tests/cases/samples/tests.py
+++ b/tests/cases/samples/tests.py
@@ -1,0 +1,65 @@
+import json
+import os
+from openpyxl import Workbook
+from restlib2.http import codes
+from ..base import AuthenticatedBaseTestCase
+from varify.samples.models import Project
+from guardian.shortcuts import assign
+
+
+class VariantUploadResourceTestCase(AuthenticatedBaseTestCase):
+    fixtures = ['initial_variants.json']
+
+    def setUp(self):
+        super(VariantUploadResourceTestCase, self).setUp()
+
+    def test_post(self):
+        book = Workbook()
+        sheet1 = book.get_active_sheet()
+        sheet1.title = 'Variants List'
+        fields = ['Chromosome', 'Start', 'Reference', 'Allele 1', 'Allele 2',
+                  'dbSNP']
+
+        # Create variants to cover all edge cases, including the case
+        # where there is a variation at the same genomic position.
+        # Also consider the case where the dbSNP id is incorrect.
+        variants = [['1', '20000', '.', 'AC', 'AC', ''],
+                    ['1', '20000', 'A', 'A', '.', ''],
+                    ['3', '20002', 'GAT', '.', '.', 'rs9160301'],
+                    ['1', '20003', '.', '.', 'TTTCTT', ''],
+                    ['3', '20004', 'A', 'C', 'C', 'rs916000'],
+                    ['1', '20007', 'GTCATTGGAACAGTC', '.',
+                     'GTCATTGGAACAGTC', '']]
+
+        # Write our fields in first.
+        for i in range(0, 6):
+            sheet1.cell(row=0, column=i).value = fields[i]
+
+        # Write the variants to the excel sheet.
+        row = 1
+        for v in variants:
+            for col in range(0, 6):
+                sheet1.cell(row=row, column=col).value = v[col]
+            row += 1
+
+        book.save('variantList.xlsx')
+
+        with open('variantList.xlsx') as fp:
+            # Assign permissions for test user to access the project.
+            p = Project.objects.get(pk=1)
+            assign('samples.view_project', self.user, p)
+
+            response = self.client.post('/api/samples/9/variants/sets/upload/',
+                                        {'name': 'variants',
+                                         'result_set_file': fp})
+            response_obj = json.loads(response.content)
+
+            # An array of matching variants are returned. Make sure
+            # that exactly 5 were found and 1 was added to the
+            # list of unmatched variants.
+            self.assertEqual(response.status_code, codes.ok)
+            self.assertEqual(response_obj['total_processed'], 6)
+            self.assertEqual(response_obj['match_count'], 5)
+            self.assertEqual(len(response_obj['unknown']), 1)
+
+        os.remove('variantList.xlsx')
--- a/varify/samples/resources.py
+++ b/varify/samples/resources.py
@@ -1,6 +1,8 @@
 import functools
+import openpyxl
 import logging
 import requests
+import itertools
 from django.core import management
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.conf.urls import patterns, url
@@ -18,6 +20,8 @@ from serrano.resources.base import ThrottledResource
 from varify.variants.resources import VariantResource
 from varify import api
 from varify.assessments.models import Assessment
+from varify.variants.models import Variant
+from varify.genome.models import Chromosome
 from .models import Sample, Result, ResultScore, ResultSet
 
 log = logging.getLogger(__name__)
@@ -390,6 +394,193 @@ class SampleResultSetResource(ThrottledResource):
         return serialize(instance, **self.template)
 
 
+class SampleResultSetUploadResource(SampleBaseResource):
+    INSERTION = 'Insertion'
+    DELETION = 'Deletion'
+
+    supported_content_types = (
+        'multipart/form-data',
+        'application/vnd.openxmlformats-officedocumentspreadsheetml.sheet')
+
+    def get_change_type(self, ref, a1, a2):
+        """
+        Given ref allele1 and allele2, returns the type of change.
+        The only case of an amino acid insertion is when the ref is
+        represented as a '.'.
+        """
+        if ref == '.':
+            return self.INSERTION
+        elif a1 == '.' or a2 == '.':
+            return self.DELETION
+
+    def is_not_found(self, request, response, pk):
+        return not self.get_object(request, pk=pk)
+
+    # Consume the variants excel file and search the database for
+    # matching queries.
+    def post(self, request, pk):
+        sample = self.get_object(request, pk=pk)
+
+        # Try opening the excel file that was uploaded.
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book,
+                                              use_iterators=True)
+        except:
+            return self.render(request, 'Could not process the file. Make '
+                               'sure that the file is a valid excel file',
+                               status=codes.unprocessable_entity)
+
+        sheet = workbook.get_active_sheet()
+
+        start_row = 0
+        fields = []
+
+        # Skip the header information of the data file (if there is any)
+        # and find the row where the data starts by finding the first instance
+        # of a column title.
+        for row in sheet.iter_rows():
+            if row[0].internal_value == 'Chromosome':
+                # Retrieve all the column titles from the sheet.
+                fields = [r.internal_value.lower() for r in row]
+                break
+            start_row += 1
+
+        allele1_index = fields.index('allele 1') \
+            if 'allele 1' in fields else None
+        allele2_index = fields.index('allele 2') \
+            if 'allele 2' in fields else None
+        dbsnp_index = fields.index('dbsnp')
+        ref_index = fields.index('reference')
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        no_matches = []
+
+        # Verison 1.6.1 of openpyxl does not support slicing thus, a workaround
+        # proposed here: https://bitbucket.org/ericgazoni/openpyxl/issue/273/allow-slicing-in-iterableworksheet # noqa
+        for row in itertools.islice(sheet.iter_rows(), start_row + 1, None):
+            # Retrieve rsid, ref, start and chr from our spreadsheet
+            # so we can retrieve objects from our database.
+            dbsnp_value = row[dbsnp_index].internal_value
+            ref_value = row[ref_index].internal_value
+            start_value = row[start_index].internal_value
+
+            # In the case that the position changes and there was no matching
+            # query. Save the original to return back to the user.
+            old_position = start_value
+
+            chr_label = int(row[chr_index].internal_value)
+
+            chr_match = Chromosome.objects.get(value=chr_label)
+
+            allele1 = row[allele1_index].internal_value \
+                if allele1_index else None
+            allele2 = row[allele2_index].internal_value \
+                if allele1_index else None
+
+            is_found = False
+
+            # Fetch the variant w/ matching dbsnp id and the
+            # result which have the variant.
+            if dbsnp_value:
+                try:
+                    result = Result.objects.get(sample=sample,
+                                                variant__rsid=dbsnp_value)
+                    matches.add(result)
+                    is_found = True
+                except Result.DoesNotExist:
+                    pass
+
+            if allele1 and not is_found:
+                # Determine the type of change. This is required to resolve
+                # the case where two variants have the same genomic start
+                # position.
+                change_type = self.get_change_type(ref_value, allele1, allele2)
+
+                if change_type:
+                    # In the case of an insertion or a deletion, the start
+                    # position is always decremented by 1 and the query is
+                    # is made using only the position and chromosome.
+                    start_value -= 1
+
+                else:
+                    # Now using the two alleles, construct the alt.
+                    if allele1 == ref_value:
+                        alt_value = allele2
+                    elif allele1 != allele2:
+                        alt_value = '{0},{1}'.format(allele1, allele2)
+                    else:
+                        alt_value = allele1
+
+                # Retrieve variant needed for our result query.
+                try:
+                    correct_variant = None
+
+                    # If there was a deletion or an insertion, make the query
+                    # using the position and the chromosome and resolve any
+                    # conflicts.
+                    if change_type:
+                        variant_match = Variant.objects.filter(
+                            pos=start_value,
+                            chr=chr_match)
+
+                        if len(variant_match) == 1:
+                            correct_variant = variant_match[0]
+
+                        elif len(variant_match) > 1:
+                            for match in variant_match:
+                                # In the case of an insertion, the variants
+                                # reference string should have increased by 1
+                                # amino acid.
+                                if change_type == self.INSERTION and \
+                                        len(match.ref) == len(ref_value) + 1:
+                                    correct_variant = match
+                                    break
+                                # In the case of a deletion (where the value
+                                # of the reference is '.'. The reference should
+                                # have a length of 1 amino acid.
+                                elif change_type == self.DELETION and \
+                                        len(match.ref) == 1:
+                                    correct_variant = match
+                                    break
+
+                    else:
+                        correct_variant = Variant.objects.get(
+                            pos=start_value,
+                            ref=ref_value,
+                            alt=alt_value,
+                            chr=chr_match)
+
+                    result = Result.objects.get(
+                        sample=sample,
+                        variant=correct_variant)
+
+                    matches.add(result)
+                    is_found = True
+                except (Variant.DoesNotExist, Result.DoesNotExist):
+                    pass
+
+            # Queries were unsuccessful so save this row to return back to the
+            # user to edit/fix them.
+            if not is_found:
+                no_matches.append({
+                    'chr_label': chr_label,
+                    'Start': old_position,
+                    'Reference': ref_value,
+                    'Allele 1': allele1,
+                    'Allele 2': allele2,
+                    'dbSNP': dbsnp_value
+                    })
+
+        return {
+            'total_processed': len(matches) + len(no_matches),
+            'match_count': len(matches),
+            'unknown': no_matches
+        }
+
+upload_resource = never_cache(SampleResultSetUploadResource())
 sample_resource = never_cache(SampleResource())
 samples_resource = never_cache(SamplesResource())
 named_sample_resource = never_cache(NamedSampleResource())
@@ -417,6 +608,10 @@ urlpatterns = patterns(
     url(r'^(?P<pk>\d+)/variants/$',
         sample_results_resource,
         name='variants'),
+
+    url(r'^(?P<pk>\d+)/variants/sets/upload/$',
+        upload_resource,
+        name='variant-set-upload'),
 
     url(r'^variants/(?P<pk>\d+)/$',
         sample_result_resource,
--- a/tests/fixtures/initial_variants.json
+++ b/tests/fixtures/initial_variants.json
@@ -1,0 +1,333 @@
+[
+    {
+        "pk": 1,
+        "model": "samples.project",
+        "fields": {"notes": null,
+            "label": "Something else",
+            "modified": "2014-02-25T14:38:56.910",
+            "name": "Default",
+            "created": "2014-02-25T14:38:56.910"}
+    },
+    {
+        "pk": 1,
+        "model": "samples.batch",
+        "fields":
+        {"count": 3,
+          "name": "locus_1",
+          "investigator": null,
+          "notes": null,
+          "created": "2014-02-25T14:39:46.700",
+          "modified": "2014-02-25T14:40:16.208",
+          "label": "locus_1",
+          "project": 1,
+          "published": true
+        }
+    },
+    {
+        "model": "genome.chromosome",
+        "pk": 1,
+        "fields": {
+            "label": "1",
+            "value": "1",
+            "order": 1,
+            "code": 1
+        }
+    },
+    {
+        "model": "genome.chromosome",
+        "pk": 3,
+        "fields": {
+            "label": "3",
+            "value": "3",
+            "order": 3,
+            "code": 3
+        }
+    },
+    {
+        "model": "genome.genotype",
+        "pk": 2,
+        "fields": {
+            "label": "Het Ref",
+            "value": "0/1",
+            "order": 2,
+            "code": 2
+        }
+    },
+    {
+        "model": "variants.varianttype",
+        "pk": 1,
+        "fields": {
+            "value": "SNP",
+            "label": "SNP",
+            "code": 1,
+            "order": 1
+        }
+     },
+    {
+        "pk": 1,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 1,
+            "pos": 19999,
+            "rsid": "rs91603",
+            "alt": "AAC",
+            "ref": "A",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "pk": 6,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 1,
+            "pos": 19999,
+            "rsid": "rs91603",
+            "alt": "T",
+            "ref": "TA",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "pk": 2,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 1,
+            "pos": 20002,
+            "rsid": "rs9160399",
+            "alt": "CTTTCTT",
+            "ref": "C",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "pk": 3,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 3,
+            "pos": 20001,
+            "rsid": "rs9160300",
+            "alt": "A",
+            "ref": "AGAT",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "pk": 11,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 3,
+            "pos": 20004,
+            "rsid": "rs916000",
+            "alt": "C",
+            "ref": "A",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "pk": 4,
+        "model": "variants.variant",
+        "fields": {
+            "articles": [],
+            "liftover": null,
+            "type": 1,
+            "chr": 3,
+            "pos": 20006,
+            "rsid": "rs9160300",
+            "alt": "CGTCATTGGAACAGTC",
+            "ref": "C,CGTTGTCATTGGAACAGTC",
+            "md5": "8d908ff5406b"
+        }
+    },
+    {
+        "model": "samples.person",
+        "pk":1,
+        "fields": {
+            "mrn": "blah",
+            "sex": "Female",
+            "created": "2014-02-25T14:39:46.700",
+            "modified": "2014-02-25T14:40:16.208"
+        }
+    },
+    {
+        "pk": 9,
+        "model": "samples.sample",
+        "fields": {
+            "count": 13,
+            "bio_sample": null,
+            "name": "NA999",
+            "created": "2014-02-25T14:39:46.766",
+            "notes": null,
+            "published": true,
+            "modified": "2014-02-25T14:40:16.191",
+            "label": "NA999",
+            "project": 1,
+            "person": 1,
+            "version": 0,
+            "vcf_colname": "78",
+            "batch": 1,
+            "phenotype_modified": null,
+            "md5": null
+        }
+    },
+    {
+        "pk": 1,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 43,
+            "variant": 1,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 147
+        }
+    },
+    {
+        "pk": 2,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 43,
+            "variant": 2,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 149
+        }
+    },
+    {
+        "pk": 3,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 883,
+            "variant": 3,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 349
+        }
+    },
+    {
+        "pk": 4,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 883,
+            "variant": 4,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 549
+        }
+    },
+    {
+        "pk": 5,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 3883,
+            "variant": 6,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 33349
+        }
+    },
+    {
+        "pk": 6,
+        "model": "samples.result",
+        "fields": {
+            "haplotype_score": null,
+            "read_depth": null,
+            "coverage_alt": null,
+            "homopolymer_run": null,
+            "sample": 9,
+            "genotype": null,
+            "quality": null,
+            "base_counts": 11883,
+            "variant": 11,
+            "mq0": null,
+            "read_pos_rank_sum": null,
+            "strand_bias": null,
+            "notes": null,
+            "created": "2014-02-25T14:40:07.630",
+            "modified": "2014-02-25T14:40:07.630",
+            "mq_rank_sum": null,
+            "baseq_rank_sum": null,
+            "raw_read_depth": 933349
+        }
+    }
+
+
+]
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -9,6 +9,7 @@ DATABASES = {
 
 INSTALLED_APPS = (
     'tests',
+    'tests.cases.samples',
     'tests.cases.south_tests',
     'tests.cases.geneset_form',
     'tests.cases.sample_load_process',


varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+            to resemble their representations in the varify database."""
+        if ref == '.':
+            return 'INSERTION'
+        elif a1 == '.' or a2 == '.':
+            return 'DELETION'
+        else:
+            return 'NOCHANGE'
+
+    # Consume the variants excel file and search the database for
+    # Matching queries
```
naegelyd: Don't capitalized continuation lines

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
 
 
+class UploadResource(ThrottledResource):
+
+    supported_content_types = ('multipart/form-data',
+                               'application/vnd.openxmlformats-officedocument'
+                               + 'spreadsheetml.sheet', )
+
+    def change(self, ref, a1, a2):
+        """ Given ref allele1 and allele2, returns the type of change.
```
naegelyd: Move the `"""` to its own line so this would be:

```
"""
Given ref allele1 and...
....
"""
```

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
         return serialize(instance, **self.template)
 
 
+class UploadResource(ThrottledResource):
+
+    supported_content_types = ('multipart/form-data',
```
naegelyd: consider changing this to:

``` python
supported_content_types = (
    'multipart/form-data',
    'application/vnd.openxmlformats-officedocumentspreadsheetml.sheet'
```

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                                    column=fields.index('dbsnp')).value
+            dbsnpValue = dbsnpValue if dbsnpValue else None
+            refValue = sheet.cell(row=row,
+                                  column=fields.index('reference')).value
+            startValue = sheet.cell(row=row,
+                                    column=fields.index('start')).value
+            chrLabel = \
+                sheet.cell(row=row,
+                           column=fields.index('chromosome')).value
+            print chrLabel
```
naegelyd: Remove print statements. If anything is truly important and these aren't meant for your own testing purposes, then change these to log at the appropriate level.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
         return serialize(instance, **self.template)
 
 
+class UploadResource(ThrottledResource):
```
naegelyd: Rename to SampleResultSetUploadResource and move to bottom with the other SampleResultSet\* resources

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+            return 'INSERTION'
+        elif a1 == '.' or a2 == '.':
+            return 'DELETION'
+        else:
+            return 'NOCHANGE'
+
+    # Consume the variants excel file and search the database for
+    # Matching queries
+    def post(self, request):
+
```
naegelyd, line 418: remove blank line

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+        # Samplename must be unique and so must the batch. The latter is not
+        # specified in a variant list excel sheet.
+        sampleName = request.POST['sampleName']
+        sample = Sample.objects.filter(name=sampleName)
+        sample = sample[0]
+
+        # Try opening the excel file that was uploaded
+        try:
+            variantBook = request.FILES['qqfile']
```
naegelyd: Is there any reasoning behind naming this `qqfile`? Consider explaining the name or renaming to something more easily understood like `result_set_file`.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+            workbook = openpyxl.load_workbook(variantBook)
+        except:
+            return HttpResponse(status=codes.unprocessable_entity)
```
naegelyd: Add a message to the response explaining the error

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+
+        sheet = workbook.get_active_sheet()
+
+        # Find the row where our useful data starts
+        startRow = 0
+        for row in range(0, len(sheet.rows)):
+            cellValue = sheet.cell(row=row, column=0).value
+            if cellValue == 'Chromosome':
+                    startRow = row
```
naegelyd: Bad indent, move 4 spaces left

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                    break
+
+        # retrive all the fields from the sheet
```
naegelyd: Capitalize the `r` in `retrieve`

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+        fields = [sheet.cell(row=startRow, column=col).value
+                  for col in range(len(sheet.columns))]
+
+        fields = [field.lower() for field in fields]
```
naegelyd: Merge the `.lower()` call into the loop above this one.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+
+        matches = set()
+        noMatches = []
+
+        for row in range(startRow + 1, len(sheet.rows)):
+            # Retrive rsid, ref, start and chr from our spreadsheet
+            # So we can retrive objects from our database
+            dbsnpValue = sheet.cell(row=row,
+                                    column=fields.index('dbsnp')).value
+            dbsnpValue = dbsnpValue if dbsnpValue else None
```
naegelyd: What is the purpose of this line? Is it meant as a way to check for bad row/column pairs?

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+            print chrLabel
+
+            # Insertion, deletion or none?
+            change_type = None
+
+            chrMatch = Chromosome.objects.get(value=chrLabel)
+
+            isFound = False
+            # Fetch the variant w/ matching dbSNP id and the
+            # Results which have the variant
```
naegelyd: don't capitalize mid-sentence

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+            if dbsnpValue:
+                try:
+                    variant = Variant.objects.get(rsid=dbsnpValue)
+                    result = Result.objects.filter(sample_id=sample.id,
```
naegelyd: The result is unique on [sample + variant](https://github.com/cbmi/varify/blob/master/varify/samples/models.py#L361) so why not use `.get(..)` here instead of `.filter(...)` and then the `if len(result) > 0:` below can be removed?

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+        else:
+            return 'NOCHANGE'
+
+    # Consume the variants excel file and search the database for
+    # Matching queries
+    def post(self, request):
+
+        # Samplename must be unique and so must the batch. The latter is not
+        # specified in a variant list excel sheet.
+        sampleName = request.POST['sampleName']
```
bruth: Underscores, no camelcase.

bruth:
Change all variable names from camelcase to using underscores.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                    # Determine the type of change. This is required to resolve
+                    # the case where two variants have the same genomic start
+                    # position.
+                    change_type = self.change(refValue, allele1, allele2)
+
+                else:
+                    # Now using the two alleles, construct the alt
+                    altValue = ""
+                    if allele1 == refValue:
+                        altValue += allele2
```
naegelyd: Just set `altValue = allele2` here. No reason for the addition

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                    else:
+                        altValue += ''.join([allele1, ",", allele2]) \
```
naegelyd: Move this logic up so that the `else` block becomes:

``` python
elif allele1 != allele2:
    altValue = '{0},{1}'.format(allele1, allele2)
else:
    altValue = allele1
```

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                # Retrive variant needed for our result query
+                try:
+                    correctVariant = ''
+
+                    # If there was a deletion or an insertion, make the query
+                    # using the position and the chromosome and resolve any
+                    # conflicts
+                    if change_type:
+                        variantMatch = \
+                            Variant.objects.filter(pos=startValue,
```
naegelyd: Move this up a line and drop arguments a line like:

``` python
variantMatch = Variant.objects.filter(
    pos=startValue, chr_id=chrMatch.id)
```

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                                    break
+
+                    else:
+                        variantMatch = Variant.objects.get(pos=startValue,
+                                                           ref=refValue,
+                                                           alt=altValue,
+                                                           chr_id=chrMatch.id)
+                        correctVariant = variantMatch
+
+                    result = \
```
naegelyd: See comment above, move this up a line and put both args on same line below to reduce indent levels

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                        Result.objects.filter(sample_id=sample.id,
```
naegelyd: Also, like noted above, Result is unique on sample + variant so this can just use `.get(...)`.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                                              variant_id=correctVariant.id)
+                    if len(result) > 0:
+                        matches.update(result)
+                        isFound = True
+                except:
+                    pass
+
+            # No method worked, failed to find result
+            if isFound is False:
```
naegelyd: `if not isFound`

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                allele1 = sheet.cell(row=row,
+                                     column=fields.index('allele 1')).value
```
naegelyd: I've seen these references throughout the code. Consider saving these before the loop to prevent constant lookups. So, you would have variables containing these values(example: `alleleOneIndex = fields.index('allele 1')).value`) and then the variables can be accessed for the column values instead of doing a lookup on `fields` every time.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                allele2 = sheet.cell(row=row,
+                                     column=fields.index('allele 2')).value
+                # No matches for these rows. Save them to display to the user
+                unMatchedFields = ['Chromosome:{0}, Start:{1}, Reference:{2}',
```
naegelyd: Why is this an array? Couldn't you just declare this as a string and then remove the join below?

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+                                   ' Allele1:{3}, Allele2:{4}, dbsnp: {5}\n']
+                unMatchedVariant = ''.join(unMatchedFields).format(chrLabel,
+                                                                   startValue,
+                                                                   refValue,
+                                                                   allele1,
+                                                                   allele2,
+                                                                   dbsnpValue)
+
+                noMatches.append(str(unMatchedVariant))
```
naegelyd: It seems impossible to me that `unMatchedVariant` could be anything other than a string at this point. Remove the `str()` wrapper here unless it serves some specific purpose I am missing.

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+
+        # Strings that will display information about the results
+        results = []
+        for result in matches:
+            detailString = ['Variant ID: {0}, Position: {1}, dbSNP: {2}\n ']
```
naegelyd: Why the array(see comment above)?

varify/samples/resources.py
```diff
@@ -410,6 +607,10 @@ def get(self, request, pk):
         samples_resource,
         name='samples'),
 
+    url(r'^uploadFile/$',
```
naegelyd: Change url to be:

``` python
url(r'^(?P<pk>\d+)/variants/sets/upload/$',
```

varify/samples/resources.py
naegelyd: This should expect the sample `pk` to be sent to the `post` method as an arguement(see comment above)

varify/samples/resources.py
```diff
@@ -390,6 +393,200 @@ def get(self, request, pk):
...
+        if ref == '.':
+            return 'INSERTION'
+        elif a1 == '.' or a2 == '.':
+            return 'DELETION'
+        else:
+            return 'NOCHANGE'
+
+    # Consume the variants excel file and search the database for
+    # Matching queries
+    def post(self, request):
```
naegelyd: Add `pk` as third argument. This needs to be made to URL change below(see comment near bottom)

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,57 @@
...
+        # Also consider the case where the dbSNP id is incorrect.
+        variants = [['1', '20000', '.', 'AC', 'AC', ''],
+                    ['1', '20000', 'A', 'A', '.', ''],
+                    ['3', '20002', 'GAT', '.', '.', 'rs9160301'],
+                    ['1', '20003', '.', '.', 'TTTCTT', ''],
+                    ['3', '20004', 'A', 'C', 'C', 'rs916000'],
+                    ['1', '20007', 'GTCATTGGAACAGTC', '.',
+                     'GTCATTGGAACAGTC', '']]
+
+        # First write the fields to our file across 4 columns
```
naegelyd: This comment is confusing. It says 4 columns but the loop below runs 6 times and there are 6 fields. Clarify or remove comment

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,57 @@
...
+        for i in range(0, 6):
+            sheet1.cell(row=0, column=i).value = fields[i]
+
+        # Write the variants to the excel sheet
+        row = 1
+        for v in variants:
+            for col in range(0, 6):
+                sheet1.cell(row=row, column=col).value = str(v[col])
```
naegelyd: These are all strings already since you wrapped them all in `'...'` when creating them in the variants arrays above. remove the `str()` wrapper

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,57 @@
...
+            row += 1
+
+        book.save('variantList.xlsx')
+
+        with open('variantList.xlsx') as fp:
+            response = self.client.post('/api/samples/9/variants/sets/upload/',
+                                        {'sampleName': 'NA999',
```
naegelyd: Is this still necessary? If not, remove. If so, use underscore naming instead of camel case.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
         return serialize(instance, **self.template)
 
 
+class SampleResultSetUploadResource(ThrottledResource):
+
```
naegelyd, line 400: Add some constants here such as:

``` python
INSERTION = 'Insertion'
DELETION = 'Deletion'
NOCHANGE = 'No Change'
```

And then, throughout the code dealing with change ttypes, you can reference these constants rather than using the strings themselves so that, if these every change, they only need to change in this single spot rather than replacing all the occurrences of the strings and worrying about case mismatches, missed occurrences, etc.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        """
+        Given ref allele1 and allele2, returns the type of change.
+        The only case of an amino acid insertion is when the ref is
+        represented as a '.'
+        In all cases of an insertion or a deletion, the genomic starting
+        position is decremented by one and the ref and alt values change
+        to resemble their representations in the varify database.
+        """
+        if ref == '.':
+            return 'INSERTION'
```
naegelyd: See note above for these strings

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                            chr_id=chr_match.id)
+
+                        correct_variant = variant_match[0]
+
+                        if len(variant_match) > 1:
+                            for match in variant_match:
+                                # In the case of an insertion, the variants
+                                # reference string should have increased by 1
+                                # amino acid.
+                                if change_type == 'INSERTION' and \
```
naegelyd: See note about change type constants and start of this class

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        sample = sample[0]
+
+        # Try opening the excel file that was uploaded
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book)
+        except:
+            log.error('Could not process the file. Make sure that the file ',
+                      'is a valid microsoft excel file.')
+            return HttpResponse(status=codes.unprocessable_entity)
```
naegelyd: The HttpResponse [allows you to set the content](https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.HttpResponse.content). Add a short message in the content kwarg briefly indicating this was an error with the excel file.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+
+        sheet = workbook.get_active_sheet()
+
+        # Find the row where our useful data starts
```
naegelyd: Expand this comment to describe what you consider to be the start of the "useful data"

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            if 'allele 2' in fields else None
+
+        matches = set()
+        noMatches = []
+
+        for row in range(start_row + 1, len(sheet.rows)):
+            # Retrive rsid, ref, start and chr from our spreadsheet
+            # So we can retrive objects from our database
+            dbsnp_value = sheet.cell(row=row,
+                                     column=fields.index('dbsnp')).value
```
naegelyd: replace this with dbsnp_, reference_, start_indexes like allele*_index above

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            cell_value = sheet.cell(row=row, column=0).value
+            if cell_value == 'Chromosome':
+                start_row = row
+                break
+
+        # Retrive all the fields from the sheet
+        fields = [sheet.cell(row=start_row, column=col).value.lower()
+                  for col in range(len(sheet.columns))]
+
+        allele1_index = fields.index('allele 1') \
```
naegelyd, line 449: Will this importer work if these indexes are missing?

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                    is_found = True
+                except:
+                    pass
+
+            if allele1_index and not is_found:
+                allele1 = sheet.cell(row=row,
+                                     column=allele1_index).value
+                allele2 = sheet.cell(row=row,
+                                     column=allele2_index).value
+
```
naegelyd, line 501: Consider moving `change_type` initialization here from comment line above

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            ref_value = sheet.cell(row=row,
+                                   column=fields.index('reference')).value
+            start_value = sheet.cell(row=row,
+                                     column=fields.index('start')).value
+            chr_label = \
+                sheet.cell(row=row,
+                           column=fields.index('chromosome')).value
+
+            # Insertion, deletion or none?
+            change_type = None
```
naegelyd: Move this closer to first reference and into a deeper scope. See comment below

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+
+            is_found = False
+            # Fetch the variant w/ matching dbsnp id and the
+            # results which have the variant
+            if dbsnp_value:
+                try:
+                    variant = Variant.objects.get(rsid=dbsnp_value)
+                    result = Result.objects.get(sample_id=sample.id,
+                                                variant_id=variant.id)
+                    matches.update([result])
```
naegelyd: remove unnecessary cast to array and use `matches.add(result)`

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            if allele1_index and not is_found:
+                allele1 = sheet.cell(row=row,
+                                     column=allele1_index).value
+                allele2 = sheet.cell(row=row,
+                                     column=allele2_index).value
+
+                if allele1 == '.' or allele2 == '.' or ref_value == '.':
+                    # Regardless of an insertion or a deletion, we see that
+                    # the start position decrements by 1. We will then query
+                    # using the position only
```
naegelyd: period at end

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                    else:
+                        alt_value = allele1
+
+                # Retrive variant needed for our result query
+                try:
+                    correct_variant = ''
+
+                    # If there was a deletion or an insertion, make the query
+                    # using the position and the chromosome and resolve any
+                    # conflicts
```
naegelyd: period at end

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                    if change_type:
```
naegelyd, line 524: Do you really need this to be an Insertion or Deletion because a value of 'NOCHANGE' will still cause this to evaluate as true. Either cahnge the comment or make this more explicit by checking that change_type is INSERTION or DELETION

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                                # In the case of an insertion, the variants
+                                # reference string should have increased by 1
+                                # amino acid.
+                                if change_type == 'INSERTION' and \
+                                        len(match.ref) == len(ref_value) + 1:
+                                    correct_variant = match
+                                    break
+                                # In the case of a deletion (where the value
+                                # of the reference is '.'. The reference should
+                                # have a value of 1 amino acid.
```
naegelyd: Do you mean `...length of 1 amino acid`?

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                        variant_id=correct_variant.id)
+
+                    matches.update([result])
+                    is_found = True
+                except:
+                    pass
+
+            # No method worked, failed to find result
+            if not is_found:
+                allele1 = sheet.cell(row=row,
```
naegelyd: These values are retrieved in [multiple if statements](https://github.com/cbmi/varify/pull/297/files#diff-2ff87413b41bd8485dc6fc5b720faaf6R489). Look them up at a broader scope so they can be referenced in both cases rather than doing multiple lookups

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                                     column=allele1_index).value
+                allele2 = sheet.cell(row=row,
+                                     column=allele2_index).value
+                # No matches for these rows. Save them to display to the user
+                unMatchedFields = ['Chromosome:{0}, Start:{1}, Reference:{2}',
```
naegelyd: Any reason for this array? Use a string

varify/samples/resources.py
hassanNS, line 449: there are two types of forms. One containing 'allele1' and 'allele2' in the fields. Another containing the genotype. The spreadsheet must have both of these fields or else lookups will be wild (only dbsnp or position and no way to resolve variants w/ same position or incorrect dbsnp). 

The code for handling the type 2 form is not included here as I haven't been able to test it on an actual form of that nature so I don't know how reliable it is.

varify/samples/resources.py
hassanNS, line 524: The 'NOCHANGE' value seems pointless so I'll get rid of it. Knowing a current variants change type allows us to make the right query as conflicts only arise in the cases of position changes due to insertion/deletion.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                            alt=alt_value,
+                            chr_id=chr_match.id)
+
+                        correct_variant = variant_match
+
+                    result = Result.objects.get(
+                        sample_id=sample.id,
+                        variant_id=correct_variant.id)
+
+                    matches.update([result])
```
naegelyd: Use .add and remove the array wrapper here. `result` is a single Result since you are using `get`

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        In all cases of an insertion or a deletion, the genomic starting
+        position is decremented by one and the ref and alt values change
+        to resemble their representations in the varify database.
+        """
+        if ref == '.':
+            return self.INSERTION
+        elif a1 == '.' or a2 == '.':
+            return self.DELETION
+        else:
+            return None
```
bruth: No need to explicitly return `None`; you can leave off the `else` since the former two return.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+
+    # Consume the variants excel file and search the database for
+    # matching queries
+    def post(self, request, pk):
+        # Samplename must be unique and so must the batch. The latter is not
```
bruth: This comment is no longer relevant since the primary key is being used.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        sample = sample[0]
+
+        # Try opening the excel file that was uploaded
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book)
+        except:
+            res = 'Could not process the file. Make sure that the file ' + \
+                'is a valid microsoft excel file.'
+            return HttpResponse(content=res,
```
bruth: Use `self.render(request, res, status=codes.unprocessable_entity)` rather than constructing an `HttpResponse` instance.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        else:
+            return None
+
+    # Consume the variants excel file and search the database for
+    # matching queries
+    def post(self, request, pk):
+        # Samplename must be unique and so must the batch. The latter is not
+        # specified in a variant list excel sheet.
+        sample = Sample.objects.filter(pk=pk)
+        sample = sample[0]
```
bruth: This will cause an index error if the sample does not exist. This resource should subclass [this class](https://github.com/cbmi/varify/blob/variant-sets/varify/samples/resources.py#L50-L71) which gives you the two methods for enforcing access to the sample and getting (and caching) the sample instance given a primary key. Once you subclass, you need to add [this `is_not_found` method](https://github.com/cbmi/varify/blob/variant-sets/varify/samples/resources.py#L82-L83) to handle 404s.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            return HttpResponse(content=res,
+                                status=codes.unprocessable_entity)
+
+        sheet = workbook.get_active_sheet()
+
+        # Skip the header information of the data file (if there is any)
+        # and find the row where the data starts by finding the first instance
+        # of a column title.
+        start_row = 0
+        for row in range(0, len(sheet.rows)):
```
bruth: Cleaner to do `for i, row in enumerate(sheet.rows)`. This gives you the row index and the row itself.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+    def post(self, request, pk):
+        # Samplename must be unique and so must the batch. The latter is not
+        # specified in a variant list excel sheet.
+        sample = Sample.objects.filter(pk=pk)
+        sample = sample[0]
+
+        # Try opening the excel file that was uploaded
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book)
```
bruth: I recommend passing `use_iterators=True` so the rows are lazily loaded since you are doing a straight iteration below.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+        dbsnp_index = fields.index('dbsnp')
+        ref_index = fields.index('reference')
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        noMatches = []
+
+        for row in range(start_row + 1, len(sheet.rows)):
+            # Retrive rsid, ref, start and chr from our spreadsheet
```
bruth: "Retrive" to "Retrieve"

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            # So we can retrive objects from our database
```
bruth: Lowercase "so" and fix spelling of "retrive" like above.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            allele2 = sheet.cell(row=row, column=allele2_index).value \
+                if allele1_index else None
+
+            is_found = False
+            # Fetch the variant w/ matching dbsnp id and the
+            # results which have the variant
+            if dbsnp_value:
+                try:
+                    variant = Variant.objects.get(rsid=dbsnp_value)
+                    result = Result.objects.get(sample_id=sample.id,
```
bruth: No need to do two queries here, simplify by doing:

``` python
result = Result.objects.get(sample=sample, variant__rsid=dbsnp_value)
```

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            if 'allele 1' in fields else None
+        allele2_index = fields.index('allele 2') \
+            if 'allele 2' in fields else None
+        dbsnp_index = fields.index('dbsnp')
+        ref_index = fields.index('reference')
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        noMatches = []
```
bruth: Remove camelcase.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+            chr_match = Chromosome.objects.get(value=chr_label)
+
+            allele1 = sheet.cell(row=row, column=allele1_index).value \
+                if allele1_index else None
+            allele2 = sheet.cell(row=row, column=allele2_index).value \
+                if allele1_index else None
+
+            is_found = False
+            # Fetch the variant w/ matching dbsnp id and the
+            # results which have the variant
```
bruth: "results" to "result"

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                    # Now using the two alleles, construct the alt
+                    alt_value = ""
+                    if allele1 == ref_value:
+                        alt_value = allele2
+                    elif allele1 != allele2:
+                        alt_value = '{0},{1}'.format(allele1, allele2)
+                    else:
+                        alt_value = allele1
+
+                # Retrive variant needed for our result query
```
bruth: Fix spelling "retrive" like above.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                try:
+                    correct_variant = ''
+
+                    # If there was a deletion or an insertion, make the query
+                    # using the position and the chromosome and resolve any
+                    # conflicts.
+                    if change_type:
+                        variant_match = Variant.objects.filter(
+                            pos=start_value,
+                            chr_id=chr_match.id)
```
bruth: You can just do `chr=chr_match`.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+                                    break
+
+                    else:
+                        variant_match = Variant.objects.get(
+                            pos=start_value,
+                            ref=ref_value,
+                            alt=alt_value,
+                            chr_id=chr_match.id)
+
+                        correct_variant = variant_match
```
bruth: Just assign the above result to `correct_variant` since `variant_match` is not being used below, rather than reassigning here.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+
+                    result = Result.objects.get(
+                        sample_id=sample.id,
```
bruth: Mentioned a couple times above, if you have the instance present, there is no need to do `foo_id=foo.id`, just do `foo=foo`. Change all places that this occurs since it is more idiomatic.

varify/samples/resources.py
```diff
@@ -390,6 +393,197 @@ def get(self, request, pk):
...
+
+                    matches.add(result)
+                    is_found = True
+                except:
+                    pass
+
+            # No method worked, failed to find result
+            if not is_found:
+                # No matches for these rows. Save them to display to the user
+                unMatchedFields = \
```
bruth: No camels.

varify/samples/resources.py
hassanNS: how would this class work with the tests?
In `post` would I just do `sample = self.get_object(request, pk=pk)` ?

bruth: Yes. As a side note, the instance is cached the first time the lookup is performed (called from `is_not_found` upstream), so this second call does result in a database hit.

hassanNS: when testing, permissions must be give to access sample.view_projects as stated here
https://pythonhosted.org/django-guardian/api/guardian.shortcuts.html

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
 
+class SampleResultSetUploadResource(SampleBaseResource):
+    INSERTION = 'Insertion'
+    DELETION = 'Deletion'
+
+    supported_content_types = (
+        'multipart/form-data',
+        'application/vnd.openxmlformats-officedocumentspreadsheetml.sheet')
+
+    def change(self, ref, a1, a2):
```
naegelyd: Consider renaming to something more descriptive like `get_change_type(..)`

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+        """
+        Given ref allele1 and allele2, returns the type of change.
+        The only case of an amino acid insertion is when the ref is
+        represented as a '.'
+        In all cases of an insertion or a deletion, the genomic starting
```
naegelyd: Remove this part of the comment. This method doesn't do the decrementing Since this method only determines the type of change, this portion of the comment can be removed since it only applies in other portions of the code.

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+        dbsnp_index = fields.index('dbsnp')
+        ref_index = fields.index('reference')
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        no_matches = []
+
+        for row in sheet.iter_rows():
+            # Skip over to our start row
```
naegelyd: Why no just set `current_row = start_row + 1` before the loop begins rather than wasting iterations incrementing the `current_row` just to get to `start_row + 1`?

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+            if allele1 and not is_found:
+                # Insertion, deletion or none?
+                change_type = self.change(ref_value, allele1, allele2)
+
+                if change_type:
+                    # Regardless of an insertion or a deletion, we see that
+                    # the start position decrements by 1. We will then query
+                    # using the position only
+                    start_value -= 1
+                    # Determine the type of change. This is required to resolve
```
naegelyd: This 3 line comments applies to nothing. Remove or move to a more applicable position, not above an empty line

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+                    if allele1 == ref_value:
+                        alt_value = allele2
+                    elif allele1 != allele2:
+                        alt_value = '{0},{1}'.format(allele1, allele2)
+                    else:
+                        alt_value = allele1
+
+                # Retrieve variant needed for our result query
+                try:
+                    correct_variant = ''
```
naegelyd: preference thing, initialize as `correct_variant = None`

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+
+                    # If there was a deletion or an insertion, make the query
+                    # using the position and the chromosome and resolve any
+                    # conflicts.
+                    if change_type:
+                        variant_match = Variant.objects.filter(
+                            pos=start_value,
+                            chr=chr_match)
+
+                        correct_variant = variant_match[0]
```
naegelyd: Is there any guarantee that a variant will match this query? If not, then use

``` python
if variant_mathc.exists():
    correct_variant = variant_match[0]
```

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+
+        sheet = workbook.get_active_sheet()
+
+        # Skip the header information of the data file (if there is any)
+        # and find the row where the data starts by finding the first instance
+        # of a column title.
+        start_row = 0
+        current_row = 0
+        fields = []
+        for row in sheet.iter_rows():
```
bruth, line 442: Add line above this for clarity.

varify/samples/resources.py
bruth: You can set the offset by passing `row_offset=start_row + 1`. See [the docs](https://pythonhosted.org/openpyxl/api.html?highlight=iter_rows#openpyxl.reader.iter_worksheet.IterableWorksheet.iter_rows)

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+                    # the start position decrements by 1. We will then query
+                    # using the position only
+                    start_value -= 1
+                    # Determine the type of change. This is required to resolve
+                    # the case where two variants have the same genomic start
+                    # position.
+
+                else:
+                    # Now using the two alleles, construct the alt
+                    alt_value = ""
```
bruth: There is no need to initialize `alt_value` since an `else` block is defined below.

varify/samples/resources.py
naegelyd: There is a bug(or its by design, there seems to be some confusion on their repo) with row_offset and what it actually means and does. See [this issue](https://bitbucket.org/ericgazoni/openpyxl/issue/273/allow-slicing-in-iterableworksheet).

varify/samples/resources.py
bruth: In fact, this does not need to be initialized since an `else` block is defined below.

varify/samples/resources.py
hassanNS: hmmm I wanted to raise an index out of bound exception in the case a matching query doesn't exist as filter() raises no exceptions like get()

varify/samples/resources.py
hassanNS: this technique is buggy in version 1.6.1 @naegelyd suggested a workaround using itertools found <a href='https://bitbucket.org/ericgazoni/openpyxl/issue/273/allow-slicing-in-iterableworksheet'>here</a>.
This method seems to be safe regardless of version

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+
+                    matches.add(result)
+                    is_found = True
+                except:
+                    pass
+
+            # No method worked, failed to find result
+            if not is_found:
+                # No matches for these rows. Save them to display to the user
+                unmatched_fields = \
```
bruth: Since this an API that returns encoded data (as supposed to markup or something human readable), it's best to return the raw data so the client can format/present it how it decides. Simple do:

``` python
no_matches.append({
    'chr_label': chr_label,
    # ...
})
```

varify/samples/resources.py
```diff
@@ -390,6 +393,195 @@ def get(self, request, pk):
...
+                                                            ref_value,
+                                                            allele1,
+                                                            allele2,
+                                                            dbsnp_value)
+
+                no_matches.append(unmatched_variant)
+
+        return {
+            'results': len(matches),
+            'noResults': no_matches
```
bruth: No camels.

bruth: Also, this response body could be a bit more informative, maybe:
- `total_processed` - this makes it known to the client that the correct number of variants were processed (trust)
- `match_count` - count of variants that matched (this arguably not needed since if there are errors it's irrelevant since it's all or nothing and if it's successfully loaded, `total_processed` is the count)
- `unknown` - array of variants that did not match (renamed from `noResults`)
- `invalid` - list of rows indices that have invalid data such as parse issues

varify/samples/resources.py
```diff
@@ -410,6 +602,10 @@ def get(self, request, pk):
         samples_resource,
         name='samples'),
 
+    url(r'^(?P<pk>\d+)/variants/sets/upload/$',
+        upload_resource,
+        name='upload'),
```
bruth: Change to `name='variant-set-upload'` since it's specific to variant sets.

varify/samples/resources.py
naegelyd: Purposefully raising the exception to alter the control flow is generally frowned upon. Maybe this is better:

``` python
if variant_match.count() == 1:
    correct_varaint = variant_mathc[0]
elif variant_match.count() > 1:
    ...
```

hassanNS: do you mean `len(variant_match) == 1 ...`? count() gets the # of a particular thing in an array right?

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+    # matching queries
+    def post(self, request, pk):
+        sample = self.get_object(request, pk=pk)
+        # Try opening the excel file that was uploaded
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book,
+                                              use_iterators=True)
+        except:
+            res = 'Could not process the file. Make sure that the file ' + \
```
naegelyd: Any reason for this assignment? Seems like this string could just be passed directly into the render call

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        no_matches = []
+
+        # Verison 1.6.1 of openpyxl does not support slicing thus, a workaround
+        # proposed here: https://bitbucket.org/ericgazoni/openpyxl/issue/273
+        # /allow-slicing-in-iterableworksheet
+        # using itertools is being used to skip to the 8th row.
```
naegelyd: Remove this portion of the comment. It is explained well enough above

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+                                # In the case of a deletion (where the value
+                                # of the reference is '.'. The reference should
+                                # have a length of 1 amino acid.
+                                elif change_type == self.DELETION and \
+                                        len(match.ref) == 1:
+                                    correct_variant = match
+                                    break
+
+                    else:
+                        correct_variant = Variant.objects.get(
```
naegelyd, line 550: There is no guarantee this `get` call will find the variant of interest. Wrap in try/except to catch Variant.DoesNotExist

hassanNS, line 550: the catch right below it catches the Variant.DoesNotExist

varify/samples/resources.py
naegelyd: I meant `count`. [`filter()` returns a queryset](https://docs.djangoproject.com/en/dev/ref/models/querysets/#filter) which [has the count method](https://docs.djangoproject.com/en/dev/ref/models/querysets/#count) which you can use to check the number of elements in the queryset.

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,62 @@
...
+                                         'result_set_file': fp})
+            response_obj = json.loads(response.content)
+            # An array of matching variants are returned. Make sure
+            # that exactly 5 were found and 1 was added to the
+            # list of unmatched variants
+            self.assertEqual(response.status_code, codes.ok)
+            self.assertEqual(response_obj['total_processed'], 6)
+            self.assertEqual(response_obj['match_count'], 5)
+            self.assertEqual(len(response_obj['unknown']), 1)
+        os.remove('variantList.xlsx')
```
naegelyd, line 65: Add blank line above

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,62 @@
...
+
+        with open('variantList.xlsx') as fp:
+            # Assign permissions for test user to access the project
+            p = Project.objects.get(pk=1)
+            assign('samples.view_project', self.user, p)
+            response = self.client.post('/api/samples/9/variants/sets/upload/',
+                                        {'name': 'variants',
+                                         'result_set_file': fp})
+            response_obj = json.loads(response.content)
+            # An array of matching variants are returned. Make sure
```
naegelyd, line 57: Add blank line above

tests/cases/samples/tests.py
```diff
@@ -0,0 +1,62 @@
...
+                sheet1.cell(row=row, column=col).value = v[col]
+            row += 1
+
+        book.save('variantList.xlsx')
+
+        with open('variantList.xlsx') as fp:
+            # Assign permissions for test user to access the project
+            p = Project.objects.get(pk=1)
+            assign('samples.view_project', self.user, p)
+            response = self.client.post('/api/samples/9/variants/sets/upload/',
```
naegelyd, line 52: Add blank line above

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+
+    supported_content_types = (
+        'multipart/form-data',
+        'application/vnd.openxmlformats-officedocumentspreadsheetml.sheet')
+
+    def get_change_type(self, ref, a1, a2):
+        """
+        Given ref allele1 and allele2, returns the type of change.
+        The only case of an amino acid insertion is when the ref is
+        represented as a '.'
```
naegelyd: Missing . at end of comment.

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+        if ref == '.':
+            return self.INSERTION
+        elif a1 == '.' or a2 == '.':
+            return self.DELETION
+
+    def is_not_found(self, request, response, pk):
+        return not self.get_object(request, pk=pk)
+
+    # Consume the variants excel file and search the database for
+    # matching queries
```
naegelyd: Missing . at end of comment. This applies some of the comments below too. Check the rest of this file for this issue.

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+    def post(self, request, pk):
+        sample = self.get_object(request, pk=pk)
+        # Try opening the excel file that was uploaded
```
naegelyd: add blank line above

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+        try:
+            variant_book = request.FILES['result_set_file']
+            workbook = openpyxl.load_workbook(variant_book,
+                                              use_iterators=True)
+        except:
+            return self.render(request, 'Could not process the file. Make ' +
```
naegelyd: Remove `+` sign, the strings will automatically be concatenated as they are within parens

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+
+        matches = set()
+        no_matches = []
+
+        # Verison 1.6.1 of openpyxl does not support slicing thus, a workaround
+        # proposed here: https://bitbucket.org/ericgazoni/openpyxl/issue/273
+        # /allow-slicing-in-iterableworksheet
+        for row in itertools.islice(sheet.iter_rows(), start_row + 1, None):
+            # Retrieve rsid, ref, start and chr from our spreadsheet
+            # so we can retrive objects from our database
```
naegelyd: fix spelling: retrive --> retrieve

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+
+            # Fetch the variant w/ matching dbsnp id and the
+            # result which have the variant
+            if dbsnp_value:
+                try:
+                    result = Result.objects.get(sample=sample,
+                                                variant__rsid=dbsnp_value)
+                    matches.add(result)
+                    is_found = True
+                except:
```
naegelyd: Specify the actual exception you want to catch here, or, if you need more than one, use a [parenthesized tuple](https://docs.python.org/2/tutorial/errors.html#handling-exceptions).

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+                            alt=alt_value,
+                            chr=chr_match)
+
+                    result = Result.objects.get(
+                        sample=sample,
+                        variant=correct_variant)
+
+                    matches.add(result)
+                    is_found = True
+                except:
```
naegelyd: See note above about except block. The types of errors you want to catch here should be explicitly listed

varify/samples/resources.py
```diff
@@ -410,6 +601,10 @@ def get(self, request, pk):
         samples_resource,
         name='samples'),
 
+    url(r'^(?P<pk>\d+)/variants/sets/upload/$',
```
naegelyd: Nit. Move this down below the `name='variant-sets'` url.

varify/samples/resources.py
```diff
@@ -390,6 +394,193 @@ def get(self, request, pk):
...
+            return self.INSERTION
+        elif a1 == '.' or a2 == '.':
+            return self.DELETION
+
+    def is_not_found(self, request, response, pk):
+        return not self.get_object(request, pk=pk)
+
+    # Consume the variants excel file and search the database for
+    # matching queries
+    def post(self, request, pk):
```
naegelyd, line 421: @bruth Any reason for this `post` method to be in a separate resource and not just be the `post` handler on the `SampleResultSetsResource` resource?

bruth, line 421: Good point, but should be conditional based on the content type. The main POST method could dispatch the request (call the method corresponding to the content type).

naegelyd, line 421: Sounds good. I will take care of this later and we can leave this PR as is.

varify/samples/resources.py
```diff
@@ -390,6 +394,194 @@ def get(self, request, pk):
...
+        dbsnp_index = fields.index('dbsnp')
+        ref_index = fields.index('reference')
+        start_index = fields.index('start')
+        chr_index = fields.index('chromosome')
+
+        matches = set()
+        no_matches = []
+
+        # Verison 1.6.1 of openpyxl does not support slicing thus, a workaround
+        # proposed here: https://bitbucket.org/ericgazoni/openpyxl/issue/273
```
naegelyd: Put this URL all on one line. If flake8 doesn't like it, add `noqa` after the URL to ignore the long line

naegelyd:
The [run-tests.sh script](https://github.com/cbmi/varify/blob/master/run-tests.sh) and the [test_suite.py script](https://github.com/cbmi/varify/blob/master/test_suite.py) need to be updated to include the new `samples` module in tests that was added as part of this PR.

hassanNS:
should the tests be excluded since they are failing due to django issues?

naegelyd:
No, this code should be tested. I'm looking into the test issues.

naegelyd:
The tests have multiple problems...disable the importer tests for now and we will create a separate issue to fix them later

PR #306, opened 2014-05-21
Rebased headless_test branch with cbmi/master

hassanNS:
Moved headless_tests on top of cbmi/master and updated the travis build file


naegelyd:
I rebased the headless_test branch, should be all good now(might need to squash some commits but we can do that later.

PR #308, opened 2014-05-22
Support variant set creation from file and from current filter set

naegelyd:
The look and feel of the modal is a bit rough but all the functionality is present.


varify/samples/forms.py
```diff
@@ -38,4 +38,11 @@ def save_m2m(self):
 
     class Meta(object):
         model = Cohort
-        fields = ('name', 'description', 'user', 'published')
```
bruth, line 41: Is there a reason the `description` field was removed here?

varify/samples/resources.py
```diff
@@ -22,11 +24,25 @@
...
 from varify.variants.models import Variant
 from varify.genome.models import Chromosome
-from .models import Sample, Result, ResultScore, ResultSet
+from .models import Sample, Result, ResultSet, ResultScore
+from .forms import ResultSetForm
 
 log = logging.getLogger(__name__)
 
 
+class DateTimeEncoder(json.JSONEncoder):
```
bruth: Django [already provides this class](https://github.com/django/django/blob/master/django/core/serializers/json.py#L84-L109)

varify/samples/resources.py
```diff
@@ -566,21 +563,98 @@ def post(self, request, pk):
...
-        }
+                    'chr': chr_label,
+                    'start': old_position,
+                    'ref': ref_value,
+                    'allele1': allele1,
+                    'allele2': allele2,
+                    'dbsnp': dbsnp_value,
+                })
+
+        instance.bulk(matches)
```
bruth, line 559: This is not conditional for invalid records?

varify/samples/resources.py
```diff
@@ -566,21 +563,98 @@ def post(self, request, pk):
...
+
+        content = serialize(instance, **self.template)
+        content['num_total_records'] = len(matches) + len(no_matches)
+        content['invalid_records'] = no_matches
+
+        return HttpResponse(status=codes.created,
```
bruth: restlib2 already [uses](https://github.com/bruth/restlib2/blob/master/restlib2/serializers/_json.py) the Django JSON encoder (referenced above), so this can be changed to the below which relies on the `supported_accept_types` to determining the output type (defaults to `application/json`)

``` python
return self.render(request, content, status=codes.created)
```

varify/samples/resources.py
```diff
@@ -566,21 +563,98 @@ def post(self, request, pk):
...
+                            content=json.dumps(content, cls=DateTimeEncoder),
+                            content_type='application/json')
+
+    def _create_from_context(self, request, instance):
+        context = self.get_context(request)
+        results = context.apply(queryset=Result.objects.all())
```
bruth: Default model tree is `Result`, so this is redundant since the queryset isn't being filtered.

varify/samples/resources.py
```diff
@@ -566,21 +563,98 @@ def post(self, request, pk):
...
+        return HttpResponse(status=codes.created, content=content,
+                            content_type='application/json')
+
+    def post(self, request, pk):
+        if 'source' in request.FILES and request.FILES['source']:
+            data = request.POST
+        else:
+            data = request.data
+
+        if hasattr(request, 'user'):
```
bruth: Assuming someone using Varify does not change which URLs are protected, this check is redundant since the resource is already protected by the django-siteauth middleware.

varify/samples/forms.py
```diff
@@ -38,4 +38,11 @@ def save_m2m(self):
...
+        fields = ('name', 'user', 'published')
+
+
+class ResultSetForm(forms.ModelForm):
+    class Meta(object):
+        model = ResultSet
+        # Exclude the results, those will be dealt with externally
+        exclude = ('results',)
```
bruth: It is best practice to exclude the `user` field from forms for models that contain them. The user can be attached manually after the form has validated prior to save. This is to prevent the case that someone submits form data already containing the user ID that is not theirs. The pattern is:

``` python
form = ResultSetForm(data)

if form.is_valid():
    instance = form.save(commit=False)
    instance.user = request.user
    instance.save()
```

varify/samples/resources.py
```diff
@@ -566,21 +563,98 @@ def post(self, request, pk):
...
+            data['user'] = request.user.id
```
bruth: [See the above comment](https://github.com/cbmi/varify/pull/308/files#discussion_r12995342) about not using the form to set the user.

varify/static/js/src/ui/modals/variant-set.js
```diff
@@ -0,0 +1,241 @@
...
+
+            if (this.ui.sourceRadios.filter(':checked').val() === 'file') {
+                var formData = new FormData();
+                var file = this.ui.fileChooser.prop('files')[0];
+
+                formData.append('source', file, file.name);
+                formData.append('name', this.ui.name.val());
+                formData.append('description', this.ui.description.val());
+
+                // I am not aware of a way to do this by passing the file
```
bruth: The options passed in during a `Collection.create` will propagate to the `Model.save` call which propagates to [`Backbone.sync`](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1174-L1259) and all ajax options are respected and not overwritten. So the below should be able to be simplified to:

``` javascript
// Pass attributes for model instance, even though the data passed
// in the options will be sent to the server. These initial attributes
// could be used for display while we are waiting for the server.
this.sample.variantSets.create({
    name: this.ui.name.val(),
    description: this.ui.description.val()
}, {
    data: formData,
    processData: false,
    contentType: false,
    success: this.onSuccess,
    error: this.onError
});
```

The `cache` parameter does not apply for non-GET/HEAD requests.

varify/static/js/src/ui/modals/variant-set.js
```diff
@@ -0,0 +1,241 @@
...
+            event.preventDefault();
+
+            // Don't let the user trigger a duplicate save request
+            this.ui.save.prop('disabled', true);
+
+            this.ui.form.hide();
+            this.ui.loading.show();
+
+            if (this.ui.sourceRadios.filter(':checked').val() === 'file') {
+                var formData = new FormData();
```
bruth, line 202: `FormData` is a relatively new API: https://developer.mozilla.org/en-US/docs/Web/API/FormData#Browser_compatibility and is not well supported it appears.

varify/samples/forms.py
naegelyd, line 41: The Cohort model does not have a `description` field. When I was importing forms in the resources, this was causing an error.

varify/samples/resources.py
naegelyd, line 559: No, @mitalia and I discussed this and we decided that we would just create the variant set regardless of whether there were mismatches or not. If we want to get fancy in the future, we could create an unpublished variant set then ask the user if they want to create with mismatches or delete it but for now, we just create the variant set and associate whatever matching results we can with it.

varify/samples/forms.py
bruth, line 41: Ah! Ok then.

varify/static/js/src/ui/modals/variant-set.js
naegelyd, line 202: From what I can tell, IE10 and all current and even much older versions of Chrome and FF support it(http://caniuse.com/fileapi). Are you aware of any reliable, cross-browser upload techniques that will also play nice with the backbone change suggested in [your comment above]? If so, I'll give that method a shot

bruth, line 202: It appears there is a [small polyfill](https://github.com/francois2metz/html5-formdata) if need be. Assuming our current users use something other than IE, we can wait until an external user requires <= IE9 support.

PR #309, opened 2014-06-03
Batch the variant data into a single request

### LINKED ISSUE #28: Batch the variant data into a single request
Currently, the preview endpoint returns the IDs of the variants matching the filter conditions. Then, for each of those IDs, a request is made to `/api/samples/variants/<ID>/` to get the result data. It may be more efficient to return the full page of variant data in a single request rather than individual requests for each variant. This will serve as a discussion area and then, if we decide to do it, a PR can be issued against this issue. While the individual requests allows the variant summary table to populated row by row as the requests come back, I think retrieving them all at once would allow for a more consistent loading indicator and would keep the table from "jumping around" as new rows are created as variants load.

ryanjohara:
Fixes #28

### DIFF
--- a/varify/samples/resources.py
+++ b/varify/samples/resources.py
@@ -17,6 +17,7 @@ from varify.variants.resources import VariantResource
 from varify import api
 from varify.assessments.models import Assessment
 from .models import Sample, Result, ResultScore
+from restlib2.http import codes
 
 log = logging.getLogger(__name__)
 
@@ -246,6 +247,28 @@ class SampleResultResource(ThrottledResource):
 
         return data
 
+    post = get
+
+
+class ResultsResource(ThrottledResource):
+    template = api.templates.SampleResultVariant
+
+    def post(self, request):
+        ids_not_found = 'ids' not in request.data
+        not_a_list = not isinstance(request.data['ids'], list)
+
+        if ids_not_found or not_a_list:
+            return self.render(
+                {'message': 'An array of "ids" is required'},
+                status=codes.unprocessable_entity)
+
+        data = []
+        resource = SampleResultResource()
+        for id in request.data['ids']:
+            data.append(resource.get(request, id))
+            
+        return data
+
 
 class PhenotypeResource(ThrottledResource):
     def get(self, request, sample_id):
@@ -327,6 +350,7 @@ samples_resource = never_cache(SamplesResource())
 named_sample_resource = never_cache(NamedSampleResource())
 sample_results_resource = never_cache(SampleResultsResource())
 sample_result_resource = never_cache(SampleResultResource())
+results_resource = never_cache(ResultsResource())
 phenotype_resource = never_cache(PhenotypeResource())
 pedigree_resource = never_cache(PedigreeResource())
 
@@ -338,6 +362,7 @@ urlpatterns = patterns(
         named_sample_resource, name='named_sample'),
     url(r'^(?P<pk>\d+)/variants/$', sample_results_resource, name='variants'),
     url(r'^variants/(?P<pk>\d+)/$', sample_result_resource, name='variant'),
+    url(r'^variants/$', results_resource, name='results_resource'),
     url(r'^(?P<sample_id>.+)/phenotypes/$', phenotype_resource,
         name='phenotype'),
     url(r'^pedigrees/(?P<year>\d+)/(?P<month>\d+)/(?P<day>\d+)/(?P<name>.+)$',
--- a/varify/static/js/src/models/result.js
+++ b/varify/static/js/src/models/result.js
@@ -44,7 +44,16 @@ define([
         }
     });
 
+    var ResultCollection = Backbone.Collection.extend({
+        model: Result,
+
+        url: function() {
+            return '' + utils.getRootUrl() + 'api/samples/variants/';
+        }
+    });
+
     return {
-        Result: Result
+        Result: Result,
+        ResultCollection: ResultCollection
     };
 });
--- a/varify/static/js/src/ui/tables/body.js
+++ b/varify/static/js/src/ui/tables/body.js
@@ -3,8 +3,9 @@
 define([
     'underscore',
     'marionette',
+    '../../models',
     './row'
-], function(_, Marionette, row) {
+], function(_, Marionette, models, row) {
 
     // Represents a "frame" of rows. The model is referenced for keeping
     // track which frame this is relative to the whole series.
@@ -13,11 +14,18 @@ define([
 
         template: function() {},
 
-        itemView: row.ResultRow,
+        initialize: function(options) {
+            this.collection = new models.ResultCollection();
+            var data = {ids: options.collection.pluck('pk')};
+            this.collection.fetch({
+                data: JSON.stringify(data),
+                type: 'POST',
+                contentType: 'application/json',
+                parse: true
+            });
+        },
 
-        itemViewOptions: function (model, index) {
-            return _.defaults({resultPk: model.get('pk')}, this.options);
-        }
+        itemView: row.ResultRow
     });
 
     return {
--- a/varify/static/js/src/ui/tables/row.js
+++ b/varify/static/js/src/ui/tables/row.js
@@ -19,27 +19,11 @@ define([
             'click': 'onClick'
         },
 
-        onClick: function(events) {
+        onClick: function() {
             c.dialogs.resultDetails.open(this, this.model);
         },
 
-        initialize: function() {
-            _.bindAll(this, 'onSync');
-
-            this.data = {};
-
-            if (!(this.data.resultPk = this.options.resultPk)) {
-                throw new Error('result pk required');
-            }
-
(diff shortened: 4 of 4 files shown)


varify/static/js/src/ui/tables/body.js
```diff
@@ -13,8 +14,35 @@ define([
 
         template: function() {},
 
+        initialize: function(options) {
+            _.bindAll(this, 'onSync');
```
naegelyd: empty line after this line

varify/static/js/src/ui/tables/body.js
```diff
@@ -13,8 +14,35 @@ define([
 
         template: function() {},
 
+        initialize: function(options) {
+            _.bindAll(this, 'onSync');    
+            this.model = new models.Results();
+            var data = {
+                ids: _.map(options.collection.models, function(model) {
```
naegelyd: Consider using [pluck](http://underscorejs.org/#pluck) instead. I think something like this would work

``` javascript
var data = {ids: _.pluck(options.collection.models, 'id')};
```

varify/static/js/src/ui/tables/row.js
```diff
@@ -35,8 +33,6 @@ define([
             this.model = new models.Result({
                 id: this.data.resultPk
             });
-
-            this.model.on('sync', this.onSync);
         },
 
         onSync: function() {
```
naegelyd: If you change this from `onSync` to `onRender`, I believe these ItemViews will only be rendered once the collection is populated above. This coincides wiht my note above.

varify/static/js/src/ui/tables/body.js
```diff
@@ -13,8 +14,35 @@ define([
...
+                data: JSON.stringify(data),
+                type: 'POST',
+                contentType: 'application/json'
+            });
+            this.model.on('sync', this.onSync);
+        },
+
         itemView: row.ResultRow,
 
+        onSync: function() {
```
naegelyd: I'm not sure I follow the reasoning behind this handler. If you remove this and change the ResultRow `onSync` to `onRender`, the CollectionView and ItemViews should take care of everything for you(I think).

varify/static/js/src/ui/tables/body.js
ryanjohara: No luck with pluck:

<code>alert(_.pluck(options.collection.models, "pk"));</code>
returned
<code>,,,,,,,,,,,,,,,,,,,</code>

<code>alert(_.map(options.collection.models, function(model) {
                return model.get('pk');
            }));</code>
returned
<code>282,283,284,285,286,287,288,289,290,291,292,293,294,295,296,297,298,299,300,301</code>

Am I missing something here?

ryanjohara: <code>options.collection.pluck('pk')</code> ended up working

varify/static/js/src/ui/tables/row.js
ryanjohara: After making these changes, the <code>onRender</code> method in <code>ResultRow</code> is only executed before the fetch call responds in <code>ResultBody</code>.  When <code>onRender</code> executes, the model only has the 'pk' key/value pair, instead of the variant data it needs to display the row.

varify/samples/resources.py
```diff
@@ -246,6 +247,21 @@ def get(self, request, pk):
 
         return data
 
+    post = get
+
+
+class SampleVariantsResource(ThrottledResource):
```
naegelyd: Let's rename this `ResultsResource` as it really doesn't depend on the sample and returns results rather than variants. This can be updated below when registering the resource and URL

varify/samples/resources.py
```diff
@@ -327,6 +343,7 @@ def get(self, request, year, month, day, name):
 named_sample_resource = never_cache(NamedSampleResource())
 sample_results_resource = never_cache(SampleResultsResource())
 sample_result_resource = never_cache(SampleResultResource())
+sample_variants_resource = never_cache(SampleVariantsResource())
```
naegelyd: See note above about name change

varify/samples/resources.py
```diff
@@ -338,6 +355,7 @@ def get(self, request, year, month, day, name):
         named_sample_resource, name='named_sample'),
     url(r'^(?P<pk>\d+)/variants/$', sample_results_resource, name='variants'),
     url(r'^variants/(?P<pk>\d+)/$', sample_result_resource, name='variant'),
+    url(r'^variants/$', sample_variants_resource, name='sample_variants'),
```
naegelyd: See note above about name change

varify/static/js/src/models/result.js
```diff
@@ -44,7 +44,16 @@ define([
         }
     });
 
+    var Results = Backbone.Collection.extend({
```
naegelyd: rename ResultCollection

ryanjohara:
Results and SampleVariantsResource have been renamed ResultCollection and ResultsResource, respectively.

varify/samples/resources.py
```diff
@@ -246,6 +247,21 @@ def get(self, request, pk):
 
         return data
 
+    post = get
+
+
+class ResultsResource(ThrottledResource):
+    supported_content_types = (
+        'application/json',)
```
bruth: This is the default, so no need to defined.

varify/samples/resources.py
```diff
@@ -246,6 +247,21 @@ def get(self, request, pk):
...
+
+    template = api.templates.SampleResultVariant
+
+    def post(self, request):
+        data = []
+        for id in request.data['ids']:
+            data.append(json.loads(SampleResultResource(request, id).content))
```
bruth: There is a lot of overhead happening here with initializing separate resources and deserializing data form the returned response. Change this to:

``` python
resource = SampleResultResource()

for id in request.data['ids']:
    data.append(resource.get(request, id))
```

varify/samples/resources.py
```diff
@@ -246,6 +247,21 @@ def get(self, request, pk):
...
+
+class ResultsResource(ThrottledResource):
+    supported_content_types = (
+        'application/json',)
+
+    template = api.templates.SampleResultVariant
+
+    def post(self, request):
+        data = []
+        for id in request.data['ids']:
```
bruth, line 267: This should be checked first to ensure the `ids` key exists otherwise it's an invalid request body.

``` python
if 'ids' not in request.data or not isinstance(request.data['ids'], list):
    return self.render({'message': 'An array of "ids" is required'},
                       status=codes.unprocessable_entity)
```

varify/samples/resources.py
```diff
@@ -246,6 +247,26 @@ def get(self, request, pk):
...
         return data
 
+    post = get
+
+
+class ResultsResource(ThrottledResource):
+    template = api.templates.SampleResultVariant
+
+    def post(self, request):
+        ids_found = 'ids' not in request.data
```
naegelyd: Shouldn't this be called `ids_not_found` to reflect its negative nature

varify/samples/resources.py
```diff
@@ -246,6 +247,26 @@ def get(self, request, pk):
...
+        not_a_list = not isinstance(request.data['ids'], list)
+        if ids_found or not_a_list:
+            msg = 'An array of "ids" is required'
+            status = codes.unprocessable_entity
+            return self.render({'message': msg}, status=status)
```
naegelyd: Remove the variable declarations and just put the message and status directly into the render method parameters

varify/samples/resources.py
```diff
@@ -246,6 +247,26 @@ def get(self, request, pk):
...
+    post = get
+
+
+class ResultsResource(ThrottledResource):
+    template = api.templates.SampleResultVariant
+
+    def post(self, request):
+        ids_found = 'ids' not in request.data
+        not_a_list = not isinstance(request.data['ids'], list)
+        if ids_found or not_a_list:
```
naegelyd: blank line above `if ...:`

varify/samples/resources.py
```diff
@@ -246,6 +247,28 @@ def get(self, request, pk):
...
+
+    def post(self, request):
+        ids_not_found = 'ids' not in request.data
+        not_a_list = not isinstance(request.data['ids'], list)
+
+        if ids_not_found or not_a_list:
+            return self.render(
+                {'message': 'An array of "ids" is required'},
+                status=codes.unprocessable_entity
+                )
```
naegelyd: Move paren to end of previous line

varify/samples/resources.py
```diff
@@ -246,6 +247,27 @@ def get(self, request, pk):
...
+        if ids_not_found or not_a_list:
+            return self.render(
+                {'message': 'An array of "ids" is required'},
+                status=codes.unprocessable_entity)
+
+        data = []
+        resource = SampleResultResource()
+        for id in request.data['ids']:
+            data.append(resource.get(request, id))
+        return data
```
naegelyd, line 270: blank line above return

PR #323, opened 2014-06-12
Add workflow for viewing a variant set and adding assessments

naegelyd:


varify/samples/resources.py
```diff
@@ -628,7 +623,30 @@ def is_not_found(self, request, response, pk):
...
         instance = self.get_object(request, pk)
-        return serialize(instance, **self.template)
+        data = serialize(instance, **self.template)
+
+        for i in range(0, len(data['results'])):
+            data['results'][i]['variant'] = VariantResource.get(
+                request, data['results'][i]['variant_id'])
+            data['results'][i].pop('variant_id')
+
+            resultId = data['results'][i]['id']
```
bruth: Remove camelcase.

varify/samples/resources.py
```diff
@@ -628,7 +623,30 @@ def is_not_found(self, request, response, pk):
 
     def get(self, request, pk):
         instance = self.get_object(request, pk)
-        return serialize(instance, **self.template)
+        data = serialize(instance, **self.template)
+
+        for i in range(0, len(data['results'])):
```
bruth: If a `range` starts at 0, it does not need to explicitly set. Change to `range(len(data['result']))`

varify/samples/resources.py
```diff
@@ -628,7 +623,30 @@ def is_not_found(self, request, response, pk):
...
+            data['results'][i]['variant'] = VariantResource.get(
+                request, data['results'][i]['variant_id'])
+            data['results'][i].pop('variant_id')
+
+            resultId = data['results'][i]['id']
+            data['results'][i]['num_assessments'] = len(
+                Assessment.objects.filter(
```
bruth: Since you are only taking the length of this query, do a `.count()` on the QuerySet to prevent returning all the data.

varify/static/js/src/ui/result-details.js
```diff
@@ -0,0 +1,199 @@
...
+            // since all the elements should now be available as all the
+            // regions have been populated and their views shown.
+            this.bindUIElements();
+
+            // Reset the row and item heights and overflow styles as they may
+            // have been toggled previously.
+            this.ui.expandableRows
+                .css('height', '' + this.maxExpandableHeight + 'px')
+                .css('overflow', 'hidden');
+            this.ui.expandCollapseLinks.text(this.showMoreText);
```
bruth, line 191: Line above for clarity.

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+        attributes: function() {
+            return {
+                'data-id': this.model.id
+            };
+        },
+
+        templateHelpers: {
+            getGeneSymbol: function() {
+                var geneSymbol = '<span class=muted>unknown gene</span>';
+                if (this.variant.unique_genes && this.variant.unique_genes.length) {    // jshint ignore:line
```
bruth, line 58: Line above for consistency with the below helpers.

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+
+                return hgvsc;
+            },
+
+            getAssessmentCount: function() {
+                if (this.num_assessments) {     // jshint ignore:line
+                    return '' + this.num_assessments + ' related assessments';      // jshint ignore:line
+                }
+
+                return '<span class=muted>This variant has not been assessed</span>';
```
bruth: Move default above `if` statement to be consistent with the above helpers (or change the above helpers to this new structure).

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+                                            'the &quot;Mother&quot; dropdown.</h5>');
+            }
+
+            if (_.isEmpty(this.model.get('father_result'))) {
+                valid = false;
+                this.ui.errorMessage.append('<h5>Please select a result from ' +
+                                            'the &quot;Father&quot; dropdown.</h5>');
+            }
+
+            if (this.model.get('sanger_requested') === undefined) {
```
bruth, line 201: Would it be safer to check if this value is a boolean type?

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+                this.$el.parent().scrollTop(0);
+            }
+
+            return valid;
+        },
+
+        onFetchError: function() {
+            this.ui.loading.hide();
+            this.ui.form.hide();
+            this.ui.errorMessage.show().html('There was an error retrieving ' +
```
bruth: Put `html` on next line which should remove the need to concatenate.

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+            else {
+                this.ui.loading.hide();
+                this.resetForm();
+            }
+        },
+
+        onSaveError: function() {
+            this.ui.errorMessage.show().html('There was an error saving the assessment.');
+            // Jump to the top to show the user the error.
+            this.$el.parent().scrollTop(0);
```
bruth, line 259: The notification already has a [fixed position](https://github.com/cbmi/cilantro/blob/master/src/scss/views/_notify.scss#L2) so this should not be necessary.

bruth, line 259: Nevermind, you are using a different element for displaying the error message.

bruth, line 259: Then again, should it use the notification system since the success handler below uses it?

bruth, line 259: When `c.notify` is called, the view [is returned](https://github.com/cbmi/cilantro/blob/master/src/js/cilantro/ui/notify.js#L114) so you can manually manipulate or dismiss it to your liking.

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+            }
+        },
+
+        // Checks the radio button with the supplied name and value(all other
+        // radios with that name are unchecked).
+        setRadioChecked: function(name, value) {
+            var checkedRadio, radios;
+
+            // Lookup all the radio buttons using the supplied name
+            radios = $('input:radio[name=' + name + ']');
```
bruth: Should this lookup be scoped to this view or the parent?

varify/static/js/src/utils.js
```diff
@@ -116,9 +116,12 @@ define([
...
          * Get the route-free URL. That is, we want to remove the route at the
          * end of the URL and be left with the root URL so we can use this to
-         * construct the result URLs later on.
+         * construct the result URLs later on. For example:
+         *
+         *  http://localhost/varify/results/  becomes  http://localhost/varify/
+         *  http://localhost/variant-sets/1   becomes  http://localhost/variant-sets/
          */
-        return window.location.href.replace(/\/[^\/]*\/$/, '/');
+        return window.location.href.replace(/\/[^\/]*\/([0-9]+)?$/, '/');
```
bruth, line 124: The example `/variant-sets/1` would result in just `/` with the regexp, not `/variant-sets/`. Is the goal to trim off the last segment with an optional numeric segment or just the last segment?

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,433 @@
...
+            this.ui.motherResult.val(this.model.get('mother_result'));
+            this.ui.fatherResult.val(this.model.get('father_result'));
+            this.ui.evidenceDetails.val(this.model.get('evidence_details'));
+
+            this.ui.form.show();
+        },
+
+        saveAssessment: function() {
+            this.model.set({
+                evidence_details: this.ui.evidenceDetails.val(),    // jshint ignore:line
```
bruth, line 285: Should be values be coerced to integers since that is what they are internally saved as? If not the change event will falsely fire since a type change occurs `1` -> `'1'`.

varify/static/js/src/utils.js
naegelyd, line 124: Sorry, the comment is wrong, this is the expected behavior. Will update comment

varify/static/js/src/ui/workflows/variant-set.js
naegelyd: Should be scoped to this view. Will update

varify/static/js/src/ui/workflows/variant-set.js
naegelyd, line 259: The error message can be quite large if multiple fields are invalid and we display other form errors in the form itself. I copied the old style of using the notify system for success but it makes no sense to use 2 different methods. I will move the success handler to use a local element.

varify/static/js/src/ui/workflows/variant-set.js
naegelyd, line 201: I think this is safe. The only way this would not be undefined is if the user clicked a radio. Since this is a fixed input(radio) field, I think we are safe just checking undefined since we know the constraints on the user's possible inputs.

varify/static/js/src/ui/workflows/variant-set.js
naegelyd, line 285: It works fine without the coercion but I see what you are saying. I will make the change in case we ever start subscribing to the change event in the future

varify/samples/resources.py
```diff
@@ -607,15 +607,10 @@ class SampleResultSetResource(SampleResultSetsResource):
...
-    def get_queryset(self, request):
-        return self.model.objects.filter(user=request.user)
-
     def get_object(self, request, pk):
         if not hasattr(request, 'instance'):
-            queryset = self.get_queryset(request)
-
             try:
-                instance = queryset.get(pk=pk)
+                instance = self.model.objects.get(pk=pk)
```
bruth, line 613: This does not enforce a user permission. Is this intentional? Anyone could supply a primary key and see the variant set even if they aren't the user.

naegelyd, line 613: Yea, this was disabled yesterday for the demo and until we can get the
permissioning sorted out. Other people should be able to see your variant
set. While this current setup is not ideal, it is better than limiting
variant sets to the user that created them otherwise it sorts of defeats
the purpose of a shared analysis on a variant set. So yes, this was
intentional but is not the long term plan.

varify/static/js/src/ui/workflows/variant-set.js
```diff
@@ -0,0 +1,437 @@
...
+                if (this.variant.effects && this.variant.effects.length) {
+                    hgvsc = this.variant.effects[0].hgvs_c || hgvsc;    // jshint ignore:line
+                }
+
+                return hgvsc;
+            },
+
+            getAssessmentCount: function() {
+                var count = '<span class=muted>This variant has not been assessed</span>';
+                if (this.num_assessments) {     // jshint ignore:line
```
bruth, line 89: Line above.

PR #327, opened 2014-06-13
second attempt to add the VCF exporter

awenocur:
The VCF exporter finally generates output that can be used in R.


bruth:
Please fix the [flake8 error](https://travis-ci.org/cbmi/varify/jobs/27588830#L1189) so the build be tested.

tests/cases/exporting/tests.py
```diff
@@ -0,0 +1,45 @@
+from django.test.utils import override_settings
+from django.test import TransactionTestCase
+from django.core.cache import cache
+from django_rq import get_worker, get_queue, get_connection
+from rq.queue import get_failed_queue
+import os
+from varify.export._vcf import VcfExporter
+import hashlib
+from django import http
+from StringIO import StringIO
```
bruth: Not important here, but just as an FYI, the `cStringIO` module is an alternate implementation (written in C) that is faster than the pure Python one you've imported.

tests/cases/exporting/tests.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # Synchronously work on queue
+        worker1 = get_worker('variants')
+        worker2 = get_worker('default')
+
+        # Work on variants...
+        worker1.work(burst=True)
+        worker2.work(burst=True)
+
+        json = '{"ranges": [{"start": 1, "end": 144000000, "chrom": "'\
```
bruth: For clarity, create a dict (formatted well) and then use the `json` module to `dumps` the data.

tests/cases/exporting/tests.py
```diff
@@ -0,0 +1,45 @@
...
+               '""1"}], "samples": ["NA12891", "NA12892", "NA12878"]}'
+        request = http.Request()
+        request._stream = StringIO(json)
+        exporter = VcfExporter()
+        buff = exporter.write(None, request = request)
```
bruth: Remove spaces around the equals sign for keyword arguments.

awenocur:
Thanks for the feedback, Byron.  I'll have a couple questions for you soon as well, about how to get the appropriate info from Serrano via the iterator; I want to fix my code a bit first.  I can easily fix all these problems except for the flake8 violation; it's inherited from the master branch in a file (varify/samples/resources.py) that this commit doesn't touch.  Should I include a fix for that file, even though it's not related to the other changes?

bruth:
That sounds good. Yea, I just noticed that the issue is in master. I will fix and push it then just rebase against master when you update your PR. Should I continue reviewing or should I wait until you are done changing things?

varify/conf/urls.py
```diff
@@ -31,6 +33,9 @@
     url(r'^sample/', TemplateView.as_view(template_name='index.html'),
         name='sample'),
 
+    url(r'^api/data/export/vcf/$',
+        csrf_exempt(VcfExporterResource()), name='vcf'),
```
bruth: I presume you exempted this for scripting purposes? Serrano [already handles that](https://github.com/cbmi/serrano/blob/master/serrano/middleware.py#L9-L13) with a custom middleware. If so, you can remove the changes in this file.

varify/export/_vcf.py
```diff
@@ -0,0 +1,286 @@
...
+
+import logging
+import vcf
+from avocado.export._base import BaseExporter
+from varify.variants.models import Variant
+from varify.samples.models import Result
+from django.db.models import Q
+import json
+from StringIO import StringIO
+import time
```
bruth: Move bare imports above all the `from .. import` statements. It's cleaner to look at. Also, there is a convention we follow to order the imports based on their _scope_, i.e. standard library first, then third-party libraries, then local packages and modules. So in this case it would look like this:

``` python
import time
import json
import logging
import vcf
from cStringIO import StringIO
from django.db.models import Q
from avocado.export._base import BaseExporter
from varify.variants.models import Variant
from varify.samples.models import Result
```

varify/export/_vcf.py
```diff
@@ -0,0 +1,286 @@
...
+    short_name = 'VCF'
+    long_name = 'Variant Call Format'
+
+    file_extension = 'vcf'
+    content_type = 'text/variant-call-format'
+
+    def write(self, iterable, buff=None, *args, **kwargs):
+        header = []
+        request = kwargs['request']
+        # descriptions of the fields currently supported by the exporter;
```
bruth: Line above comment.

varify/export/_vcf.py
```diff
@@ -0,0 +1,286 @@
...
+class VcfExporter(BaseExporter):
+    short_name = 'VCF'
+    long_name = 'Variant Call Format'
+
+    file_extension = 'vcf'
+    content_type = 'text/variant-call-format'
+
+    def write(self, iterable, buff=None, *args, **kwargs):
+        header = []
+        request = kwargs['request']
```
bruth: It should not be assumed the `request` is present. The exporter should handle the case when it's not available.

awenocur:
I think it's safe to continue reviewing, just with the understanding that I plan on the exporter behaving in essentially the same manner for both GET and POST requests.  I'll be moving all that VCF generating code out of the big if block and get rid of the else clause in _vcf.py.  For a GET request, I think it should do what it does now for POST, just getting the result IDs from the iterator rather than looking them up itself.

varify/export/_vcf.py
```diff
@@ -0,0 +1,286 @@
...
+        # descriptions of the fields currently supported by the exporter;
+        # this is to be prepended to the actual header, describing lines
+        VcfFileHeader =\
```
bruth: variables names should be lowercase and using underscores rather than camelcase. Only class names use the former format.

bruth: Use multiple-line string plus `textwrap.dedent` to remove the leading whitespace:

``` python
# import above..
from textwrap import dedent

vcf_file_header = dedent("""\
...
""")  # noqa
```

varify/export/_vcf.py
```diff
@@ -0,0 +1,286 @@
...
+            '##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">\n'\
+            '##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic '\
+            'depths for the ref and alt alleles in the order listed">\n'\
+            '##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate'\
+            ' read depth (reads with MQ=255 or with bad mates are '\
+            'filtered)">\n'\
+            '##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype '\
+            'Quality">\n'\
+            '#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT'
+        buff = self.get_file_obj(buff)
```
bruth, line 84: Line above for clarity.

varify/conf/urls.py
awenocur: I exempted this, because the csrf check fails when a tool other than a Web browser invokes the exporter with a POST request.  How recently was that feature added to token authentication?  I added that under an older version of Serrano.

bruth: 4 months ago https://github.com/cbmi/serrano/commit/56aba3ab7a23dd8bf322a9d577fa64e686dfc9ef

It appears the middleware is specified after the CSRF middleware in the settings file which is why it’s not being respected. I will move that as well with the flake8 fix.

bruth: I pushed the commits to fix the mentioned issues. The next time you push make sure you rebase against master first.

varify/export/_vcf.py
awenocur: That makes sense.  Once I have the VCF generating code out of that if block, I guess the exporter should check for the presence of a request first, then determine whether there be appropriate POST data.  If those conditions aren't met, it should rely on the iterator to get the list of results.

varify/export/_vcf.py
awenocur: Do you know of any examples where this is used for long lines?  Is it acceptable to break the 80 char rule for this?

bruth: Yes, text blocks can be ignored. I've updated the code example above with a trailing comment that is used flake8 to ignore lines (the comment itself must be two spaces after the end of the line for it to be valid).

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
+# VCF exporter
+# The primary purpose of this exporter is not to generate verbatim copies of
+# VCF entries loaded into the db, but to use VCF as a common format that
+# integrates into a bioinformatics workflow.  This exporter works with a client
+# currently hosted at http://github.research.chop.edu/wenocur/varify_client
```
naegelyd: Move this whole comment block down [to the class](https://github.com/cbmi/varify/pull/327/files#diff-8ced1d27820ada2af0d1ee56249e360cR21) it is describing.

naegelyd: When you say

> This exporter works with a client currently hosted at ...

Do you mean that the exported VCF will not work anything other than the client being linked to? That is, does the exporter put anything into the VCF that would require it to be opened with your client or could any normal client capable of importing VCFs process and VCF exported from this exporter.

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+
+class VcfExporter(BaseExporter):
+    short_name = 'VCF'
+    long_name = 'Variant Call Format'
+
+    file_extension = 'vcf'
+    content_type = 'text/variant-call-format'
+
+    def write(self, iterable, buff=None, request=None, *args, **kwargs):
+
```
naegelyd: Remove blank link

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+        # figure out what we call this data source:
```
naegelyd: Capitalize first letter of comment

``` python
# Figure out what ... 
```

naegelyd: Remove trailing `:`

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+            ##source=''' + vcf_source + '''
+            ##reference=GRCh37
+            ##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
+            ##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">
+            ##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth (reads with MQ=255 or with bad mates are filtered)">
+            ##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
+            #CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT''')  # noqa
+        buff = self.get_file_obj(buff)
+
+        # this shall contain sample labels if POST data are detected
```
naegelyd: See comment above about capitalizing first letter. Also, add periods at the ends of the sentences. This applies to most of the block comments in this code.

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+        ending_bp = []
+
+        # array to manage project permissions from command-line utility
+        permitted_projects = None
+
+        # POST should ignore the iterable by design, since it interfaces with
+        # a dedicated client
+        # decode the parameters passed from the client
+        if request:
+            if request.method == 'POST':
```
naegelyd: This nesting is not necessary, change this to:

``` python
if request and request.methods == 'POST':
    ...
```

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+                data = json.load(request._stream)
+                # get the list of sample labels
+                labels = data['samples']
+                # dictionaries encoded by the client to store chromosome ranges
+                range_dicts = None
+
+                # chromosome ranges are optional
+                if 'ranges' in data:
+                    range_dicts = data['ranges']
+                # TODO: catch invalid/missing entries in the following dicts:
```
naegelyd: Remove TODO, implement the catching for invalid/missing entries.

varify/export/_vcf.py
```diff
@@ -0,0 +1,303 @@
...
+            # arrays; store the predicate as a Q object
+            for chr, start, end in zip(chromosomes, beginning_bp, ending_bp):
+                next_criterion = Q(variant__chr__label=chr) & Q(
+                    variant__pos__lt=end + 1) & Q(
+                    variant__pos__gt=start - 1)
+                if range_criteria is None:
+                    range_criteria = next_criterion
+                else:
+                    range_criteria |= next_criterion
+        else:  # the iterable is being used
```
naegelyd: Blank line above

varify/export/_vcf.py
awenocur: The exporter generates a conventional VCF file.  It can work with either the iterator passed from Serrano, or a client that sends a POST request using JSON.  While this is designed for use with a particular client, anything that sends the appropriate HTTP request can invoke the exporter.
The designated client just spits out the VCF file verbatim.

varify/export/_vcf.py
awenocur: Where should I be adding/removing a blank line here?

naegelyd: Above the `else` in line 130

test_suite.py
```diff
@@ -13,6 +13,7 @@
         'sample_load_process',
         'south_tests',
         'geneset_form',
+        'exporting',
```
naegelyd: Nit, rename this to `export` to match the module it is testing

awenocur:
I'm removing the line where the exporter implicitly closes the file descriptor associated with buff.  This is causing the test to fail when the exporter generates buff as a cStringIO object (an inherited behavior in the absence of a buff kwarg).  Does Serrano expect the returned buffer to be associated with an open file descriptor, and attempt close it when necessary?

tests/cases/export/tests.py
```diff
@@ -0,0 +1,53 @@
...
+    def setUp(self):
+        cache.clear()
+        get_queue('variants').empty()
+        get_queue('default').empty()
+        get_failed_queue(get_connection()).empty()
+
+@override_settings(VARIFY_SAMPLE_DIRS=SAMPLE_DIRS)
+class SampleLoadTestCase(QueueTestCase):
+    def test_pipeline(self):
+
```
naegelyd: Remove blank line

tests/cases/export/tests.py
```diff
@@ -0,0 +1,53 @@
...
+        # Work on variants...
+        worker1.work(burst=True)
+        worker2.work(burst=True)
+
+        test_params = {'ranges': [{'start': 1, 'end': 144000000, 'chrom': 1}],
+                       'samples': ['NA12891', 'NA12892', 'NA12878']}
+
+        json = dumps(test_params)
+        # create a nonsensical request just so that the exporter can function
```
naegelyd: Capitalize first letter in comment and add period at the end

tests/cases/export/tests.py
```diff
@@ -0,0 +1,53 @@
...
+        request = http.HttpRequest()
+        request._stream = StringIO(json)
+        request.META['SERVER_NAME'] = 'test'
+        request.META['SERVER_PORT'] = 0
+        request.method = 'POST'
+        exporter = VcfExporter()
```
naegelyd: Blank line above

tests/cases/export/tests.py
```diff
@@ -0,0 +1,53 @@
...
+        buff = exporter.write(None, request = request)
+        hash = hashlib.md5(buff.getvalue()[-500:])
+        buff.close()
+        self.assertEqual(hash.hexdigest(), '990fe158f2e2390a8c3bd0d673ed40d1')
```
naegelyd: Blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+from avocado.export._base import BaseExporter
+from varify.samples.models import Result, Project, Sample
+
+log = logging.getLogger(__name__)
+
+# The following generates a varg list for Python 2.7, that triggers an
+# exception in the Python 2.6 standard library.
+sys_major_version = version_info[0]
+sys_minor_version = version_info[1]
+if(sys_major_version < 3 and sys_minor_version < 7):
```
naegelyd: Use more succinct tuple comparison. Example [here](https://github.com/cbmi/avocado/blob/fd02e174d990237dbc87e129a3ae47263065ec9a/setup.py#L12).

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+    unicode_conv_vargs = {}
+else:
+    unicode_conv_vargs = {'errors': 'backslashreplace'}
+
+# VCF exporter
```
naegelyd: Remove the line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+# The primary purpose of this exporter is not to generate verbatim copies of
```
naegelyd: Move this comment block(lines 25-28) under the class definition(line 31)

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def write(self, iterable, buff=None, request=None, *args, **kwargs):
+        # Figure out what we call this data source.
+        vcf_source = gethostname()
+        if request:
+            vcf_source = request.get_host()
+
+        # These are descriptions of the fields currently supported by the
+        # exporter.
+        # This is to be prepended to the actual header, describing lines.
```
naegelyd: Merge this comment with line above.

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+            ##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">
+            ##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth (reads with MQ=255 or with bad mates are filtered)">
+            ##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
+            #CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT''')  # noqa
+
+        buff = self.get_file_obj(buff)
+
+        # This shall contain sample labels if POST data are detected.
+        labels = None
+        # These shall contain ranges if POST data are detected.
```
naegelyd, line 89: Blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+        beginning_bp = []
+        ending_bp = []
+
+        # This is an array to manage project permissions from the command line
+        # utility.
+        permitted_projects = None
+
+        # POST should ignore the iterable by design, since it interfaces with
+        # a dedicated client.
+        # Decode the parameters passed from the client.
```
naegelyd: Merge this line with comment on line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+        if request and request.method == 'POST':
+            data = json.load(request._stream)
+            # Get the list of sample labels.
```
naegelyd, line 103: blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+            labels = data['samples']
+            # These are dictionaries encoded by the client to store chromosome
```
naegelyd, line 106: blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+            # ranges.
+            range_dicts = None
+
+            # Chromosome ranges are optional.
+            if 'ranges' in data:
+                range_dicts = data['ranges']
+            if range_dicts:
```
naegelyd, line 114: blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+                    permitted_projects.append(sample.project.id)
+                # No projects were specified on the command line, so there
+                # should be only one permitted.
+                assert len(permitted_projects) <= 1
+
+        # The following is an ORM-based implementation that works for now;
+        # this should be migrated to use Avocado if possible.
+        # Start with a QuerySet for all results.
+        all_results = Result.objects.get_query_set()
+        # These shall be Q objects.
```
naegelyd, line 152: Blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+                    filter(label__in=labels).distinct('project')
+                for sample in get_permitted_project_samples:
+                    permitted_projects.append(sample.project.id)
+                # No projects were specified on the command line, so there
+                # should be only one permitted.
+                assert len(permitted_projects) <= 1
+
+        # The following is an ORM-based implementation that works for now;
+        # this should be migrated to use Avocado if possible.
+        # Start with a QuerySet for all results.
```
naegelyd: Merge this comment with comment on line above

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
+from django.http import QueryDict
+from serrano.resources.exporter import ExporterResource
+from avocado.models import DataView
+
+
+class VcfExporterResource(ExporterResource):
+
```
naegelyd, line 9: Remove blank line above

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
...
+        # TODO: remove following hack once Serrano ceases to require data attr
+        # the following is a hack, to prevent Serrano from crashing
+        request.data = QueryDict('')
+        return super(VcfExporterResource, self).post(
+            request, 'vcf', **kwargs)
+
+    def dispatch(self, *args, **kwargs):
+        return super(VcfExporterResource, self).dispatch(*args, **kwargs)
+
+    supported_accept_types = ('*/*',)
```
naegelyd: Move these lines(32-34) above the functions right below the class definition(line 6)

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+                        vcf.model._Call(
+                            next_row, next_label,
+                            row_call_format(None, None, None, None)))
+
+            # The following code really should be part of PyVCF:
+            # sorting the calls within the row:
+            reordered_samples = [None] * len(next_row.samples)
+            for call in next_row.samples:
+                index = next_row._sample_indexes[call.sample]
+                # the following line checks for an exceptional condition
```
naegelyd: Add blank line above and capitalize first letter of ocmment

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+                # this should be handled in a later version of Varify
+                # rather than being thrown, and should not be added
+                # to PyVCF
+                assert reordered_samples[index] is None
+                reordered_samples[index] = call
+            next_row.samples = reordered_samples
```
naegelyd, line 361: Add blank line above

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+
+            writer.write_record(next_row)
+
```
naegelyd, line 364: ~~Does `writer` need to be closed?~~

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+        selected_results = all_results.prefetch_related(
+            'sample', 'variant').prefetch_related(
+            'variant__chr').filter(
+            label_criteria).filter(
+            range_criteria).order_by(
+            'variant__chr__order', 'variant__pos')
+
+        # Ensure results are from a particular project.
+        if permitted_projects:
+            selected_results = \
```
naegelyd: Change this to the following for readbility

``` python
selected_results = selected_results.filter(
    sample__project__id__in=permitted_projects)
```

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+                if range_criteria is None:
+                    range_criteria = next_criterion
+                else:
+                    range_criteria |= next_criterion
+
+        else:  # The iterable is being used.
+            result_ids = []
+            for row in iterable:
+                result_ids.append(row[0])
+            # Here, range_criteria actually matches the results from the
```
naegelyd, line 183: Add blank line above

varify/export/_vcf.py
awenocur, line 364: Either the writer itself, or the file it encapsulates needs to be closed.  If I close it inside the exporter, it results in a useless return value under some circumstances.  See my other comment/question about Serrano's defined behavior when calling exporters.

varify/export/_vcf.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # If no ranges were provided, or the iterator is being used,
+        # use a dummy Q object.
+        if range_criteria is None:
+            range_criteria = Q()
+
+        # Grab the results, finally; take the intersection of the two Q
+        # objects defined above, sort by the order defined in the VCF v4
+        # specification.
+        selected_results = all_results.prefetch_related(
```
naegelyd: This line is very difficult to read, this can be broken up like this and it will be more readable IMO:

``` python
selected_results = all_results.prefetch_related('sample', 'variant')\
    .prefetch_related('variant__chr')\
    .filter(label_criteria)\
    .filter(range_criteria)\
    .order_by('variant__chr__order', 'variant__pos')
```

Also, the `label_criteria` and `range_criteria` can be combined into a single `Q` object so that we only need a single filter call here instead of 2 calls for each `label_criteria` and `range_criteria`.

tests/cases/export/tests.py
```diff
@@ -0,0 +1,71 @@
...
+from django.test import TransactionTestCase
+from django.core.cache import cache
+from django_rq import get_worker, get_queue, get_connection
+from rq.queue import get_failed_queue
+from varify.export._vcf import VcfExporter
+
+TESTS_DIR = os.path.join(os.path.dirname(__file__), '../..')
+SAMPLE_DIRS = [os.path.join(TESTS_DIR, 'samples')]
+
+class QueueTestCase(TransactionTestCase):
```
naegelyd: ~~You can import QueueTestCase from [this test](https://github.com/cbmi/varify/blob/52323813b9ecbfe21959fa68866fcdd758fe9456/tests/cases/sample_load_process/tests.py#L21) to avoid recreating the same code here. Ideally this should be moved out into the base case but for now you can just import it from there until I move it over.~~

naegelyd: I moved it to base and updated the references, will be pushing as soon as the local tests finish. Please remove this and import `QueueTestCase` from its new home in `tests.cases.base` once I update master with the changes. Just add

``` python
from ..base import QueueTestCase
```

to your imports.

tests/cases/export/tests.py
```diff
@@ -0,0 +1,71 @@
...
+        # Work on variants...
+        worker1.work(burst=True)
+        worker2.work(burst=True)
+
+        exporter = VcfExporter()
+
+        # This first test is for 3 samples, with a range on one chromosome.
+        test_params = {'ranges': [{'start': 1, 'end': 144000000, 'chrom': 1}],
+                       'samples': ['NA12891', 'NA12892', 'NA12878']}
+        request = SampleLoadTestCase.prepare_request(test_params)
```
naegelyd: Is there any reason you are preparing the request yourself and not using the built in Django test client? Serrano has an example of making a request for a file [here](https://github.com/cbmi/serrano/blob/cb022657f6e9a364e73f4a45631e192becc3e46f/tests/cases/resources/tests/exporter.py#L56) and there is an example of sending parameters in a post request [here](https://github.com/cbmi/serrano/blob/830889abeb38efd131cd1c26312d73c0e49eabc8/tests/cases/resources/tests/context.py#L52-L69). By using the client, you shouldn't need to explicitly call write and you can test the full stack in that it will make sure all the URLs match and the request is routed to your exporter correctly and the correct file and file type is returned.

tests/cases/export/tests.py
```diff
@@ -0,0 +1,71 @@
...
+        buff = exporter.write(None, request = request)
+        hash = hashlib.md5(buff.getvalue()[-500:])
+        buff.close()
+
+        self.assertEqual(hash.hexdigest(), '990fe158f2e2390a8c3bd0d673ed40d1')
```
naegelyd: Would it be safer to check the line count here rather than hashing and getting the hex digest? The number of lines in the header should be constant, call them `H` lines. Then, you should have one line per variant right? If so, then the total number of lines in the returned file should be `H + V` where `V` is the number of expected variants. You can get that number by running a query matching your `test_params` so something like:

``` python
variant_count = Variant.objects.filter(....).count()
```

And then assert that `variant_count` is the same as the number of lines in the exporter VCF plus some `H` lines to account for the header VCF lines.

tests/cases/export/tests.py
```diff
@@ -0,0 +1,71 @@
...
+        # This second test is for 2 samples, to check that the remaining one
+        # is indeed being excluded.
+        test_params = {'ranges': [{'start': 1, 'end': 144000000, 'chrom': 1}],
+                       'samples': ['NA12891', 'NA12892']}
+        request = SampleLoadTestCase.prepare_request(test_params)
+        buff = exporter.write(None, request = request)
+        hash = hashlib.md5(buff.getvalue()[-500:])
+        buff.close()
+
+        self.assertEqual(hash.hexdigest(), '50dd077b5c8f55366f625e44beebf203')
```
naegelyd: See comment above about checking line count instead of hash.

tests/cases/export/tests.py
naegelyd: The reason I suggest the above approach is it makes the test independent of the sample files. If we were to change the sample files or the library changes and the lines are output in a different order in the file then the hash will be different and the test will fail even though things are still OK.

varify/conf/urls.py
```diff
@@ -31,6 +32,8 @@
     url(r'^sample/', TemplateView.as_view(template_name='index.html'),
         name='sample'),
 
+    url(r'^api/data/export/vcf/$', VcfExporterResource(), name='vcf'),
```
naegelyd, line 35: You should not need to manually register this URL. When you register the exporter in `varify/export/__init__.py` this endpoint should automatically be connected to the VcfExporter via [this resource](https://github.com/cbmi/serrano/blob/830889abeb38efd131cd1c26312d73c0e49eabc8/serrano/resources/exporter.py).

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+# The following generates a varg list for Python 2.7, that triggers an
+# exception in the Python 2.6 standard library.
+if version_info < (2, 7):
+    unicode_conv_vargs = {}
+else:
+    unicode_conv_vargs = {'errors': 'backslashreplace'}
+
+
+class VcfExporter(BaseExporter):
+    # VCF exporter
```
naegelyd, line 51: Remove this line(line 24).

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+        vcf_file_header = textwrap.dedent('''\
+            ##fileformat=VCFv4.1
+            ##fileDate= ''' + time.strftime("%Y%m%d") + '''
+            ##source=''' + vcf_source + '''
+            ##reference=GRCh37
+            ##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
+            ##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">
+            ##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth (reads with MQ=255 or with bad mates are filtered)">
+            ##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
+            #CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT''')  # noqa
```
naegelyd, line 82: Could you do this instead and remove the need for the textwrap library:

``` python
vcf_file_header = """##fileformat=VCFv4.1
##fileDate={0}   
##source={1} 
##reference=GRCh37
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">
##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth (reads with MQ=255 or with bad mates are filtered)">
##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype Quality">
#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT
""".format(time.strftime('%Y%m%d'), vcf_source)
```

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
...
+    def is_not_found(self, request, response, *args, **kwargs):
+        return super(VcfExporterResource, self).is_not_found(
+            request, response, 'vcf', **kwargs)
+
+    def get(self, request, **kwargs):
+        return super(VcfExporterResource, self).get(
+            request, 'vcf', **kwargs)
+
+    def post(self, request, **kwargs):
+        # TODO: remove following hack once Serrano ceases to require data attr
```
naegelyd, line 21: Is there a bug report filed in Serrano for this? If not, please create one and I will update Serrano if there is an issue there.

awenocur, line 21: This seems no longer to cause a crash, since I changed the client to use JSON rather than plain text, however I'll submit a bug report nonetheless.
However when I remove the overridden method, there's an unrelated runtime error; weird.

tests/cases/export/tests.py
awenocur: I chose to do a hash, because the content of the lines should be tested too.  The tests should check that the samples queried are indeed represented in the output in the proper order, and that all relevant fields are being generated properly.  Testing it programmatically in a manner independent of the test data would be a significant undertaking, as I'd have to create a second VCF importer (or exporter) independent of the existing one.
If the order of the lines changes, the test should break.  The exporter is meant to sort variants.

awenocur: I could add a management command to spit out the test results and their checksums so that it would be easy to update the test.

naegelyd: Leave it as is, I didn't realize the exporter was putting them in any special order.

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # This shall contain sample labels if POST data are detected.
+        labels = None
+
+        # These shall contain ranges if POST data are detected.
+        chromosomes = []
+        beginning_bp = []
+        ending_bp = []
+
+        # This is an array to manage project permissions from the command line
```
naegelyd, line 94: I'm not sure I understand the reference to the `command line utility` here. The exporter should be able to be used like all the other ones and not require a specific command line utility.  Can you clarify this comment? Also, shouldn't the permitted projects be determined on the basis of the `user` making the request?

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+
+            else:
+                get_permitted_project_samples = \
+                    Sample.objects. \
+                    filter(label__in=labels).distinct('project')
+                for sample in get_permitted_project_samples:
+                    permitted_projects.append(sample.project.id)
+                # No projects were specified on the command line, so there
+                # should be only one permitted.
+                assert len(permitted_projects) <= 1
```
naegelyd, line 145: I don't think raising an `AssertionError` is the right thing to do here. I think there could probably be a better exception raise here with a descriptive message. If this assertion fails, what is returned as the response to the client?

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # The following is an ORM-based implementation that works for now;
+        # this should be migrated to use Avocado if possible. Start with a
+        # QuerySet for all results.
+        all_results = Result.objects.get_query_set()
```
naegelyd, line 150: Add blank line below

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+        # This is a lookup dict for sample indexes; this is linked to each
+        # PyVCF Record object.
+        sample_indexes = {}
+
+        # Keep track of the number of samples detected.
+        sample_num = 0
+
+        # Loop over all Results returned.
+        for result in selected_results:
+
```
naegelyd: Remove blank line

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+        row_call_format._types.append('String')
+        row_call_format._types.append('Integer')
+        row_call_format._types.append('Integer')
+
+        # This is a lookup dict for sample indexes; this is linked to each
+        # PyVCF Record object.
+        sample_indexes = {}
+
+        # Keep track of the number of samples detected.
+        sample_num = 0
```
naegelyd, line 236: Couldn't you just do 

``` python
sample_num = selected_results.values_list('sample__label', flat=True).distinct().count()
```

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+                # Make it known that this variant has a PyVCF record.
+                rows[variant.id] = next_row
+
+                # The order of rows as retrieved from the DB should be
+                # right for the VCF.
+                ordered_rows.append(next_row)
+
+            # This is a hack to replace NULLs in the DB with zero where
+            # appropriate.
+            ref_coverage = 0
```
naegelyd: Could you just use `ref_coverage = result.coverage_ref or 0`?

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+            if result.coverage_ref:
+                ref_coverage = result.coverage_ref
+            alt_coverage = 0
```
naegelyd: Same as above

varify/export/_vcf.py
```diff
@@ -0,0 +1,337 @@
...
+        # template.
+        fake_template_file = StringIO(vcf_file_header +
+                                      template_sample_string)
+        template_reader = vcf.Reader(fake_template_file)
+        writer = vcf.Writer(buff, template_reader)
+        fake_template_file.close()
+
+        # Add nulls to replace missing calls; this is necessary for
+        # variants not called for all samples in the VCF; this should
+        # really be done by PyVCF.
```
naegelyd, line 336: Consider filing an issue(ideally with PR) on PyVCF if this belongs in their code.

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
...
+    supported_accept_types = ('*/*',)
+    supported_content_types = ('application/json',)
+
+    def get_view(self, request, attrs=None):
+        # Just return the variant ID, we will do the lookup ourselves. It's
+        # easier to access the data in the exporter than to try to hard code
+        # all the concept IDs here.
+        return DataView(json='{"columns": [31]}')
+
+    def is_not_found(self, request, response, *args, **kwargs):
```
naegelyd, line 16: Is this necessary? The `ExporterResource` [already defines this](https://github.com/cbmi/serrano/blob/master/serrano/resources/exporter.py#L127-L129) and passes the export_type which should be `vcf`.

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
+from django.http import QueryDict
+from serrano.resources.exporter import ExporterResource
+from avocado.models import DataView
+
+
+class VcfExporterResource(ExporterResource):
+    supported_accept_types = ('*/*',)
```
naegelyd, line 7: Is this safe? Wouldn't this accept just about anything? Might want to limit to the types you are actually willing to accept.

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
...
+            request, 'vcf', **kwargs)
+
+    def post(self, request, **kwargs):
+        # TODO: remove following hack once Serrano ceases to require data attr
+        # the following is a hack, to prevent Serrano from crashing
+        request.data = QueryDict('')
+        return super(VcfExporterResource, self).post(
+            request, 'vcf', **kwargs)
+
+    def dispatch(self, *args, **kwargs):
```
naegelyd: Is this necessary? Seems do to nothing but call the same method on the super class

varify/export/resources.py
```diff
@@ -0,0 +1,34 @@
...
+        return super(VcfExporterResource, self).dispatch(*args, **kwargs)
+
+    attrs = ''
```
naegelyd: I'm probably missing something but I don't see the reason for this line. Can you explain it? Either way, move this under this line above

``` python
supported_content_types = ('application/json',)
```

varify/export/_vcf.py
awenocur, line 94: The permitted projects should be limited based on the user, and on whether that user is explicitly allowing multiple projects.  By default, if a query spans multiple projects it should fail, to prevent unintentional violations of an IRB.

varify/export/_vcf.py
awenocur, line 236: I think the end result would be the same, but sample_num would not have the same value within the loop.  It's incremented to generate default indexes for samples, which may or may not later be replaced by indexes supplied by the user.

varify/export/_vcf.py
awenocur: I think that would be equivalent.

varify/export/_vcf.py
awenocur, line 336: I plan on doing so, eventually.  I've seen some issues within the past 6 months pertaining to how the writer API should be enhanced, so I should try using the PyVCF HEAD before writing any issues myself.  Unfortunately, it's been awhile since the last stable release, and I doubt a new one is coming anytime soon, so I think we'll have to live with a workaround for awhile.

varify/export/resources.py
awenocur, line 7: Yeah, that was a hack while I was experimenting with POST content types; good catch.

varify/export/resources.py
awenocur, line 16: ~~I followed the example from the review exporter here, just to get a list of IDs from Serrano.  It's quite possible this isn't necessary.~~

whoops; looking at the wrong line; this might be an artifact of the exporter being originally written for an old version of Harvest

varify/export/resources.py
awenocur: I think that's another thing I had in there to help with debugging; it probably could be removed safely.

varify/export/_vcf.py
awenocur, line 94: I think I found a good example to follow to get a list of permitted projects by user.  [This list](https://github.com/cbmi/varify/blob/master/varify/variants/resources.py#L174-179) should be intersected with the one currently used.

varify/export/resources.py
awenocur: This looks like something that was necessary in the past.  I added it awhile ago to make Serrano accept the old plain text content.  It seems to have no effect now.

varify/export/_vcf.py
awenocur, line 145: Currently it just fails with a 500 error.  I added that as a quick way to make the exporter fail by design when necessary.  I guess it ultimately should trigger a 403 error somehow.  Is there an exception it can throw to do that, or should the exporter generate a 403 on its own?

naegelyd:
@awenocur It seems that this exporter class servers two purposes. The first is supporting export to VCF given a certain set of parameters in the POST data. The second is to export in the standard way using an iterable(this is what all the other exporters do). I recommend removing all the POST data export related code and making this exporter follow the same standard as all the other exporters. I am not saying the POST data export code does not add value, I just think it should be in another endpoint and that this VCF exporter should follow the same conventions as the other exporters. Does it make sense to you to move the POST data export code to a different endpoint and keep the ...`/export/vcf/` endpoint for the standard exporter that supports the iterable?

awenocur:
It does make sense to break this up into two endpoints.  Are there any examples of a non-standard API endpoint like this?  There should also be a module that the two can import for shared code, which ATM is most of it.

naegelyd:
I suppose the [phenotype and pedigree resource](https://github.com/cbmi/varify/blob/31125826a47df5e78636c46be18c9af2feac9a05/varify/samples/resources.py#L270-L343) are a bit non-standard. They might be a good example to follow or at least they are closest thing I can think of at the moment.

varify/export/_vcf.py
awenocur, line 145: found it: http://www.django-rest-framework.org/api-guide/exceptions#permissiondenied

varify/conf/urls.py
awenocur, line 35: Is there some way I can make the ExporterRootResource generate an appropriate DataView for the VCF exporter?  When I have the URL generated automatically, it sends an iterator with the default DataView.  I was following the example for the Review exporter, which also requires a custom view.

naegelyd:
@awenocur Please rebase this on `master` as some significant changes have been made to `master`.

naegelyd:
@awenocur The tests were failing because the effects were written in different order in different cases [here](https://github.com/cbmi/varify/pull/327/files#diff-8ced1d27820ada2af0d1ee56249e360cR27). Changing the above line to:

``` python
allEffects = variant.effects.all().order_by('gene__symbol')
```

_should_ solve the issue and make the effects string deterministic again unless we run into the situation where effects share a gene symbol. We could try sorting by `id` but I believe that is what got us into this situation. This goes back to the larger issue [commented on above](https://github.com/cbmi/varify/pull/327#discussion_r14898964) of comparing hashes for the files. The hash is extremely fragile as we saw here where the content of the file was exactly the same but the effect ordering was different. A more robust way of checking would be to take the VCF data from the response content and load it into PyVCF or some other parser and make sure that all the rows are present and correct. This would remove any dependency on ordering of effects or the like and strictly validate the information content and not so much of the file structure and ordering.

Anyway, this will fix the tests but I would at least consider making the tests more flexible by using some other validation mechanism other than hash comparisons to avoid any more of these types of errors in the future. I am not sure my proposal above is the best but I feel there must be a better way of validating than the hash.

PR #335, opened 2014-06-19
Fix broken paging

### LINKED ISSUE #326: Paging broken
Commit [9fb150b41c307c9d84d558afa13ac9c8129ca861](https://github.com/cbmi/varify/commit/9fb150b41c307c9d84d558afa13ac9c8129ca861) seems to have broken paging.

ryanjohara:
Fixes #326

### DIFF
--- a/varify/samples/resources.py
+++ b/varify/samples/resources.py
@@ -254,19 +254,16 @@ class ResultsResource(ThrottledResource):
     template = api.templates.SampleResultVariant
 
     def post(self, request):
-        ids_not_found = 'ids' not in request.data
-        not_a_list = not isinstance(request.data['ids'], list)
-
-        if ids_not_found or not_a_list:
-            return self.render(
-                {'message': 'An array of "ids" is required'},
-                status=codes.unprocessable_entity)
+        if (not request.data.get('ids') or
+                not isinstance(request.data['ids'], list)):
+            return HttpResponse(status=codes.unprocessable_entity,
+                                content='Array of "ids" is required')
 
         data = []
         resource = SampleResultResource()
         for id in request.data['ids']:
             data.append(resource.get(request, id))
-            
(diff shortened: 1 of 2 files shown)


varify/samples/resources.py
```diff
@@ -257,16 +257,15 @@ def post(self, request):
         ids_not_found = 'ids' not in request.data
         not_a_list = not isinstance(request.data['ids'], list)
 
-        if ids_not_found or not_a_list:
-            return self.render(
-                {'message': 'An array of "ids" is required'},
-                status=codes.unprocessable_entity)
+        if ids_not_found or not_a_list or len(request.data['ids']) == 0:
```
bruth: It would be cleaner to just do:

``` python
if not request.get('ids') or not isinstance(request.data['ids'], list):
    # ...
```

rather than initializing the variables above. Just as an FYI, calling `dict.get(..)` defaults to `None` if the key is not present.

bruth:
Please sign off the commit. That is why the Travis error occurred.

varify/samples/resources.py
```diff
@@ -254,19 +254,17 @@ class ResultsResource(ThrottledResource):
...
-        ids_not_found = 'ids' not in request.data
-        not_a_list = not isinstance(request.data['ids'], list)
-
-        if ids_not_found or not_a_list:
-            return self.render(
-                {'message': 'An array of "ids" is required'},
-                status=codes.unprocessable_entity)
+        if (not request.data.get('ids') or
+            not isinstance(request.data['ids'], list) or
+                len(request.data['ids']) == 0):
```
bruth: This condition is not needed since an empty list is falsy which is asserted in the first condition.

ryanjohara:
The flake8 issues should be resolved now.

PR #336, opened 2014-06-23
Reduce result template to include only data used by client

### LINKED ISSUE #239: Reduce result template to include only the data used by the client
I believe that there are a number of fields that are unused by the client. These fields should not be included in the returned data in order to cut down on the size of the data. Marking as `needinfo` originally because I have not identified the unused fields yet.

ryanjohara:
Remove following fields from result.js:
baseq_rank_sum
genotype_quality
haplotype_score
quality_by_depth
sample.batch
sample.count
variant.effects.codon_change
variant.effects.description
variant.effects.functional_class
variant.effects.region
variant.effects.transcript.gene.chr
variant.effects.transcript.gene.families
variant.effects.transcript.gene.id
variant.effects.transcript.gene.name
Fixes #239

### DIFF
--- a/varify/api/templates.py
+++ b/varify/api/templates.py
@@ -95,9 +95,8 @@ GeneSearch = {
 
 # Limited gene information to be embedded in a Transcript
 TranscriptGene = {
-    'exclude': ['synonyms', 'transcripts'],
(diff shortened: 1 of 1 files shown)


ryanjohara:
After adding the genotype field back to the SampleResult template (since it is in fact used by the client), I successfully compared screenshots of the results page and the variant details modal before and after this change.

PR #343, opened 2014-07-01
Replace variant resource SolveBio queries with JS dashboard code

davecap:
SolveBio API queries are proxied through a Django view in Varify (`/api/solvebio/`).
The variant details modal loads a local iframe (`/api/solvebio/dashboards/variant-detail`) which is an AngularJS app that queries the local SolveBio API proxy for data, and presents the results in ClinVar and Medline widgets. 
Queries are initiated when the variant detail modal loads. Also, the SolveBio query from the variant API resource was removed (speeds up the results page).

Questions:
- What is the recommended way to make loading `solvebio.js` optional?
- Are you ok with an external S3 dependency, or would you rather bundle our JS code within the app?


davecap:
Not really ready to be merged yet, I'm mostly looking for feedback on this. Thanks!

naegelyd:
I will try to take a look at this today. I am OK with having an external dependency. We already load a couple JS files from an external CDN so loading yours from S3 should be fine. I am not sure about the best way to make the loading optional. @bruth should have a better idea of how to do that.

bruth:
@davecap If the solvebio.js script is wrapped a module definition, then it's easier to do conditional loading. You can see how to support various environments [here](https://github.com/umdjs/umd). For us, the script needs to at least support AMD, but you could have it support vanilla JavaScript and even Node if you desire.

davecap:
@bruth thanks for the tip, I'll add AMD support then. With the AMD support, does it matter at all if it's hosted on S3 or locally?

bruth:
No it does not. Most loaders have support for mapping a module name (solvebio) to a full URL for non-local modules. See http://requirejs.org/docs/api.html#config-paths as an example.

davecap:
Will submit a new pull request.

PR #353, opened 2014-07-30
Update varify to leverage varify-data-warehouse

naegelyd:


tests/settings.py
```diff
@@ -9,27 +9,32 @@
...
-    'tests.cases.commands',
+    'tests.cases.resources',
 
     'varify',
 
-    'varify.raw',
-    'varify.raw.sources',
-
     'varify.assessments',
+    'varify.genes',
```
bruth: Apps that have the same last import name (e.g. `genes`) will cause all but the first to be acknowledged. I see that with `assessments`, `genes`, `phenotypes`, `samples`, `genome`, and `variants`.

bruth:
@naegelyd It appears some of the tests [have errored](https://travis-ci.org/cbmi/varify/jobs/31355843#L1133). That may be why the coverage drop 10%.

naegelyd:
@bruth I assume you are talking about the `RequestException`s, `SSLError`s, etc. Those are all part of the gene rank test case and are intentionally triggered and expected in the test itself. All 11 tests run successfully. The reason for the decrease in coverage is a result of my reshuffling of some code and tests between varify and vdw and force pushing the changes onto this branch.

bruth:
The errors are value-related, e.g. `ValueError: invalid literal for int() with base 10: 'BAD_RANK'`

naegelyd:
All those errors are expected. See [here](https://github.com/cbmi/varify/blob/master/tests/cases/commands/tests.py#L483-L506) for the `ValueError` generating test case.

bruth:
Ok.

PR #361, opened 2014-09-08
Don't hide loading overlay until variant data is returned

### LINKED ISSUE #358: Loading indicator on results page is cleared before results are displayed
The loading indicator on the results page appears to be cleared after the call to `api/data/preview/` returns but that call does not actually return the results, it only returns a set of IDs of interest. Those IDs are then used to retrieve the actual records with a secondary call to `api/samples/variants/`. The loading message should not be cleared until this second call returns successfully.

naegelyd:
Fix #358.

This fixes a bug where the loading overlay was being hidden by Cilantro
as soon as the call to /api/data/preview/ returned which was problematic
in Varify because the variant data is not retrieved by that call but
instead comes back in a successive call to /api/samples/variants/. The
loading overlay is now hidden after that second call.

### DIFF
--- a/varify/static/js/src/ui/tables/body.js
+++ b/varify/static/js/src/ui/tables/body.js
@@ -24,11 +24,16 @@ define([
         },
 
         _fetch: function() {
+            var _this = this;
+
             this.collection.fetch({
                 data: JSON.stringify({ids: this._collection.pluck('pk')}),
                 type: 'POST',
                 contentType: 'application/json',
-                parse: true
+                parse: true,
+                success: function() {
+                    _this.trigger('synced');
+                }
             });
         },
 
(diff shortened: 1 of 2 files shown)


varify/static/js/src/ui/tables/body.js
```diff
@@ -28,7 +29,10 @@ define([
                 data: JSON.stringify({ids: this._collection.pluck('pk')}),
                 type: 'POST',
                 contentType: 'application/json',
-                parse: true
+                parse: true,
+                success: function() {
+                    c.trigger('body:collection:synced');
```
bruth: I think it is more appropriate for the table view to fire this event and simply denote that it is `synced` or `read`, rather than the global Cilantro object. The results workflow below can then listen to the table view directly.

PR #378, opened 2014-10-01
Allow users to select multiple samples

### LINKED ISSUE #374: Add multi-sample selection to the sample chooser modal

naegelyd:
Fix #374.


varify/static/js/src/ui/modals/sample.js
```diff
@@ -268,34 +275,70 @@ define([
...
+                html = 'Current: <strong>' + models[0].get('label') +
                     '</strong> from project <strong>' +
-                    model.get('project') + ' (' + model.get('batch') + ')</strong>';
-
+                    models[0].get('project') + ' (' + models[0].get('batch') +
+                    ')</strong>';
             }
             else {
-                html = 'Please select a sample.';
+                html = '<strong>' + models.length + '</strong> samples selected.';
```
bruth: Do you think it's worth listing the sample labels here rather than just the count? The list could be quite long for finding which are selected. Also I can't imagine more than a few (?) samples would be selected at once.

naegelyd: I went back and forth on this and then figured that they could get the sample labels from the filter display in the context panel if they need to see them. I was trying to avoid adding more display elements and further extending the dimensions of the modal by trying to display what could be a very long list of samples depending on who is using this and what they are attempting. I will add the sample labels since we're bound to get a request from a user soon enough...

PR #382, opened 2014-10-02
Limit cohort formatter to cohorts from projects user has permision for

### LINKED ISSUE #376: Cohort formatter does not restrict cohort list when request user is present
The cohort formatter contains [a method](https://github.com/cbmi/varify/blob/61f488722ea2b5d3fa6ba8296a99b49aa2abeebe/varify/samples/formatters.py#L56) for retrieving the cohort list but it does not limit this based on the requesting users projects and permissions. The cohort list should be limited like it is done [here](https://github.com/cbmi/varify/blob/2f7a764a1c25bdde263ea9fa38c2e3eadbccc3ae/varify/variants/resources.py#L65-L72).

naegelyd:
Fix #376.

### DIFF
--- a/varify/samples/formatters.py
+++ b/varify/samples/formatters.py
@@ -2,10 +2,11 @@ try:
     from collections import OrderedDict
 except ImportError:
     from ordereddict import OrderedDict
+from django.conf import settings
 from django.db.models import Q
+from guardian.shortcuts import get_objects_for_user
 from avocado.formatters import registry as formatters
 from serrano.formatters import HTMLFormatter
-from django.conf import settings
 
 
 class AlamutFormatter(HTMLFormatter):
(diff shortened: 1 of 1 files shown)


varify/samples/formatters.py
```diff
@@ -60,13 +61,21 @@ def _get_cohorts(self, **context):
...
+            # All cohorts this user has permission to.
+            projects = None
             if 'request' in context:
                 perms |= Q(user=context['request'].user)
+                projects = get_objects_for_user(
+                    context['request'].user, 'samples.view_project')
+
+            cohorts = Cohort.objects.filter(perms, batch=None)
+
+            if projects:
```
bruth: Use `if projects is not None`, otherwise the queryset returned from line 68 will be evaluated.

PR #383, opened 2014-10-03
Wait until samples sync to set sample label in results workflow

naegelyd:
Fix case where the result page would hang crash because the sample could not be found by ID because the sample collection hadn't synced yet.


varify/static/js/src/ui/workflows/results.js
```diff
@@ -20,13 +20,25 @@ define([
...
+
+            this.renderLabel();
+        },
+
+        renderLabel: function() {
             var samples = utils.samplesInContext(this.data.context, this.data.samples);
 
-            numbers.renderCount(this.ui.count, count);
-            if (samples.length === 1) {
+            if (samples.length === 0) {
```
bruth, line 36: In theory this should never occur correct? Since the results do not show unless a sample is chosen?

naegelyd, line 36: It is possible. Even if you have selected a sample(or samples) that will be in `this.data.context` but there is no guarantee that `this.data.samples` has synced yet. In apps(some Varify instances) with many samples, the context can be synced before the sample and that is why we re-render when the samples return. This is usually only an issue when people directly navigate to the results page before opening another page first but that is why I added this 0 length check.

PR #384, opened 2014-10-03
Embed a SolveBio variant-detail dashboard

davecap:
Bundles in the SolveBio JS library which loads the dashboard iframe from an external cookie-less domain (AWS S3).
Remove SolveBio requests from variant resource.
Using the new dashboard view requires the solvebio python module >= 1.5.2 and app credentials.
The JS view obtains a SolveBio OAuth2 token from a view within the django_solvebio app (part of the solvebio module).


naegelyd:
Sorry for the late reply on this. I like the changes being proposed in this PR but have delayed merging because of the look of the variant details modal now. When SolveBio is enabled, the UI in the variant details modal includes the UI changes from this PR and then, at the very bottom, is the current varify details UI. The reason I haven't merged this yet that I worry users will wonder where the variant details went and won't want to scroll all the way down to see them. I think we need to consider what we can do to either get all the relevant details integrated into the top of the solve bio API or provide some quick way of accessing the current details UI elements without scrolling past the articles first. While the changes in the PR will be very useful to users, I worry that those who are just looking for some quick information(say on cohort frequencies) will need to scroll past all the SolveBio content to look at what they are after.

@davecap Do those concerns make sense to you? What do you think we can do to feature more of the traditional variant details content towards the top(or more quickly accessible) while still keeping your additions from this PR?

davecap:
@naegelyd Thanks for the review. I agree that it makes sense to swap the order of the content (put the current variant details above the dashboard component). It is our goal to eventually serve up all the same information, but for now that's probably the best fix.

By the way, is there a way to get the genome build for a variant (from the variant resource)? Right now we hard-code the genome build, but that's probably not correct.

naegelyd:
Hmm, I'm not sure but I will look into it. @bruth might know.

bruth:
@davecap We are currently on hg19 for all variants. We are not exposing it right now, but that is a quick change to do so.

davecap:
I think it would be helpful to expose it. We're using the suggested sample data which seems to be in hg18.

leipzig:
i have hg19 sample data ready. will issue a pr today

PR #386, opened 2014-10-07
Set sample from context when absent in workspace

### LINKED ISSUE #385: Workspace doesn't refresh when navigated to from another page
Steps to recreate:
1. Navigate directly to the workspace page. At this point, you will see the variant set and sample details rendered normally.
2. Create a variant set if one doesn't already exist.
3. Click on a variant set and you will navigate to the variant set page.
4. Click on the workspace link in the top navbar.

After `4` you would expect the same view as in `1` but that isn't the case because the `select` handler isn't triggered like it is when the whole site first loads.

naegelyd:
Fix #385.


varify/static/js/src/utils.js
```diff
@@ -281,6 +281,25 @@ define([
...
+    var sampleIdsInContext = function(context) {
+        /*
+         * Utility method for retrieving the full list of sample IDs in
+         * the current context.
+         */
+        var sampleIds = [], json;
+
+        if (context && (json = context.get('json'))) {
+            _.each(json.children, function(child) {
+                if (child.concept && child.concept === 2) {
```
bruth: The first condition isn't necessary since the second is strict equality check.

varify/static/js/src/utils.js
```diff
@@ -281,6 +281,25 @@ define([
...
+                    sampleIds = child.value;
+                }
+            });
+        }
+
+        return sampleIds;
```
bruth, line 305: I recall the prior PRs handling a single ID vs. a list. Will the sample IDs be converted by this point? i.e Is this guaranteed to be a list?

naegelyd, line 305: Those recent changes [always set the value as a list in the context](https://github.com/cbmi/varify/blob/9094abbe86fc2b6260720711baa66c6e124b7ab7/varify/static/js/src/ui/modals/sample.js#L359-L364). I suppose someone who hasn't logged in for quite some time might have a "legacy" context that would not be in list form. I will update this to make sure the returned value is a list.

